PR #1, opened 2019-03-18
Add web module boilerplate to simplify the HTML setup

simonprev:
- `config/test.exs` changes were made to prevent this error:

```elixir
** (ArgumentError) cookie store expects conn.secret_key_base to be at least 64 bytes
```

used when fetching session in the new `browser` pipeline.

- The `{Credo.Check.Refactor.MapInto, false},` is to prevent Elixir 1.8 warning 😉
- The `ElixirBoilerplateWeb.Errors.Helpers` module is used in forms generated by `phx.gen.html` so it is useful to have it. (It is also used in forms we create from scratch so it’s in every projects)

The final result for a fresh clone will be:

image

### DIFF
--- a/lib/elixir_boilerplate_web/home/templates/index.html.eex
+++ b/lib/elixir_boilerplate_web/home/templates/index.html.eex
@@ -1,0 +1,10 @@
+<div align="center">
+  <br />
+  <img src="https://user-images.githubusercontent.com/11348/52080254-520cb580-2565-11e9-8c21-156cf0b7bcf3.png" width="500" />
+  <p><br />This repository is the stable base upon which we build our Elixir projects at Mirego.<br />We want to share it with the world so you can build awesome Elixir applications too.</p>
+
+  <div>
+    <strong>Health:</strong>
+    <%= @health %>
+  </div>
+</div>
--- a/lib/elixir_boilerplate_web/layouts/templates/app.html.eex
+++ b/lib/elixir_boilerplate_web/layouts/templates/app.html.eex
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8"/>
+    <meta http-equiv="X-UA-Compatible" content="IE=edge"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
+    <link rel="stylesheet" href="<%= Routes.static_path(@conn, "/css/app.css") %>"/>
+    <%= csrf_meta_tag() %>
+  </head>
+
+  <body>
+    <main role="main">
+      <p role="alert"><%= get_flash(@conn, :info) %></p>
+      <p role="alert"><%= get_flash(@conn, :error) %></p>
+
+      <%= render @view_module, @view_template, assigns %>
+    </main>
+
+    <script type="text/javascript" src="<%= Routes.static_path(@conn, "/js/app.js") %>"></script>
+  </body>
+</html>
--- a/lib/elixir_boilerplate_web/layouts/templates/app.html.eex
+++ b/lib/elixir_boilerplate_web/layouts/templates/app.html.eex
@@ -1,10 +1,10 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    <meta charset="utf-8"/>
-    <meta http-equiv="X-UA-Compatible" content="IE=edge"/>
-    <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
-    <link rel="stylesheet" href="<%= Routes.static_path(@conn, "/css/app.css") %>"/>
+    <meta charset="utf-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <link rel="stylesheet" href="<%= Routes.static_url(@conn, "/css/app.css") %>" />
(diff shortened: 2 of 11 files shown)


lib/elixir_boilerplate_web/layouts/templates/app.html.eex
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8"/>
+    <meta http-equiv="X-UA-Compatible" content="IE=edge"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
+    <link rel="stylesheet" href="<%= Routes.static_path(@conn, "/css/app.css") %>"/>
```
remiprev:
```suggestion
    <link rel="stylesheet" href="<%= Routes.static_url(@conn, "/css/app.css") %>"/>
```

lib/elixir_boilerplate_web/layouts/templates/app.html.eex
```diff
@@ -0,0 +1,21 @@
...
+
+  <body>
+    <main role="main">
+      <p role="alert"><%= get_flash(@conn, :info) %></p>
+      <p role="alert"><%= get_flash(@conn, :error) %></p>
+
+      <%= render @view_module, @view_template, assigns %>
+    </main>
+
+    <script type="text/javascript" src="<%= Routes.static_path(@conn, "/js/app.js") %>"></script>
```
remiprev:
```suggestion
    <script type="text/javascript" src="<%= Routes.static_url(@conn, "/js/app.js") %>"></script>
```

lib/elixir_boilerplate_web/layouts/templates/app.html.eex
```diff
@@ -0,0 +1,21 @@
...
+    <link rel="stylesheet" href="<%= Routes.static_path(@conn, "/css/app.css") %>"/>
+    <%= csrf_meta_tag() %>
+  </head>
+
+  <body>
+    <main role="main">
+      <p role="alert"><%= get_flash(@conn, :info) %></p>
+      <p role="alert"><%= get_flash(@conn, :error) %></p>
+
+      <%= render @view_module, @view_template, assigns %>
```
remiprev: If we want to be consistent with other function calls 😄 

```suggestion
      <%= render(@view_module, @view_template, assigns) %>
```

lib/elixir_boilerplate_web/layouts/templates/app.html.eex
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8"/>
```
remiprev: I think we usually try to keep the space between the last `"` and `/>`? Or did we stop doing that?

lib/elixir_boilerplate_web/home/templates/index.html.eex
```diff
@@ -0,0 +1,10 @@
+<div align="center">
+  <br />
+  <img src="https://user-images.githubusercontent.com/11348/52080254-520cb580-2565-11e9-8c21-156cf0b7bcf3.png" width="500" />
+  <p><br />This repository is the stable base upon which we build our Elixir projects at Mirego.<br />We want to share it with the world so you can build awesome Elixir applications too.</p>
```
gingman: The first `<br />` could be outside the paragraph

lib/elixir_boilerplate_web/router.ex
```diff
@@ -7,9 +7,24 @@ defmodule ElixirBoilerplateWeb.Router do
...
+    plug(:protect_from_forgery)
+    plug(:put_secure_browser_headers)
+    plug(:put_layout, {ElixirBoilerplateWeb.Layouts.View, :app})
+  end
+
   scope "/", ElixirBoilerplateWeb do
     pipe_through(:api)
 
-    get("/health", Health.Controller, :index)
+    get("/health", Health.Controller, :index, as: :api_health)
```
gcauchon: Why are we customizing the _name helpers_ for both routes?

lib/elixir_boilerplate_web/home/templates/index.html.eex
simonprev: No the space is not the same if the `br` is not inside a `p` tag

lib/elixir_boilerplate_web/router.ex
simonprev: Ah! I forgot to mention it in the PR description but this is a issue we have on all of our projects:

Since we use `Health.Controller` instead of `HealthController`, Phoenix can’t guess the name of the route.

Without the `:as` option:

```sh
_path  GET  /health  ElixirBoilerplateWeb.Health.Controller :index
_path  GET  /        ElixirBoilerplateWeb.Home.Controller :index 
```

With the `:as` option:

```sh
api_health_path  GET  /health  ElixirBoilerplateWeb.Health.Controller :index
      home_path  GET  /        ElixirBoilerplateWeb.Home.Controller :index
```

Maybe this could be a PR in Phoenix to warn or raise if we try to generate an empty path helper. Anyway, we need to generate valid helpers 😄

remiprev: Great explanation! I think we could avoid prefixing these routes with `api_` though. What do you think?

```suggestion
    get("/health", Health.Controller, :index, as: :health)
```

PR #4, opened 2019-03-21
Add empty dummy data setup

remiprev:
The empty file content is pretty much the same as `seeds.exs`.

And by running it in our CI script, we can make sure that our _dummy data_ scripts will always be up-to-date 🎉

### DIFF
--- a/priv/repo/dummy.exs
+++ b/priv/repo/dummy.exs
@@ -1,0 +1,5 @@
+# Script for populating the database with disposable, development-oriented data. You can run it as:
+#
+#     mix run priv/repo/dummy.exs
+#
+# We recommend using test factories here, since they already provide good sample data.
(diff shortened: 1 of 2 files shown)


priv/repo/dummy.exs
```diff
@@ -0,0 +1,11 @@
+# Script for populating the database with disposable, development-oriented data. You can run it as:
+#
+#     mix run priv/repo/dummy.exs
+#
+# Inside the script, you can read and write to any of your
+# repositories directly:
+#
+#     ElixirBoilerplate.Repo.insert!(%ElixirBoilerplate.SomeSchema{})
```
garno: Should we add a note stating that we recommend the usage of factories with `ex_machina`? I don’t recall any dummy data using `insert!` nor `update!`.

remiprev: Yes, good idea. Done!

PR #8, opened 2019-03-27
Remove non-existent file from coveralls configuration

garno:
We noticed in a private project that this file isn’t present in the codebase. Therefore, there is no need to skip it when running coveralls.

### DIFF
--- a/coveralls.json
+++ b/coveralls.json
@@ -3,7 +3,8 @@
     "minimum_coverage": 30
   },
   "skip_files": [
-    "elixir_boilerplate/release/migration.ex",
+    "lib/elixir-boilerplate/release_tasks.ex",
+    "lib/mix/tasks/erlang.check_version.ex",
     "test/support"
   ]
 }


coveralls.json
```diff
@@ -3,7 +3,6 @@
     "minimum_coverage": 30
   },
   "skip_files": [
-    "elixir_boilerplate/release/migration.ex",
```
remiprev: Were we supposed to ignore [this file](https://github.com/mirego/elixir-boilerplate/blob/master/lib/elixir_boilerplate/release_tasks.ex) instead?

garno: I’m not familiar with the history of this project so I can’t tell if it has just been renamed. Should I add it to the ignore list?

garno: I feel like it is something we might want to test, no?

Boubalou: Technically, this file contains the migration process when deploying an OTP release, which is used by this script: 

https://github.com/mirego/elixir-boilerplate/blob/master/rel/commands/migrate.sh

I feel like it must be ignored, indeed.

Boubalou: And yes, it should be the file proposed by @remiprev that is added there as `skip_files`.

garno: I’ve added `mix/tasks/erlang.check_version.ex` as well.

coveralls.json
```diff
@@ -3,7 +3,8 @@
...
+    "elixir-boilerplate/release_tasks.ex",
+    "mix/tasks/erlang.check_version.ex",
     "test/support"
```
remiprev: If we have `test/support` here, shouldn’t the two other paths begin with `lib`? 🤔

garno: I had the same thought and it seems like `lib/` is optional. I’ll add it anyway to remove some magic.

PR #15, opened 2019-04-04
Do not freeze patch versions for Elixir, Erlang, Node.js and NPM

### LINKED ISSUE #13: Strict version requirements
`make dependencies` fails saying `You are using Node.js 10.15.3 but the required version specified in package.json is 10.14.2`

If this is not a **strict** requirements, [package.json](https://github.com/mirego/elixir-boilerplate/blob/master/assets/package.json#L4) could be changed to `>=10.14.2` (I can make this simple change and send a PR, but want to understand if it would be OK)

remiprev:
Instead of freezing whole version numbers for Erlang, Elixir, Node.js and NPM, we now only freeze major and minor numbers. So now the version requirements are truly:

- Node.js (`~> 10.14`)
- NPM (`~> 6.9`)
- Elixir (`~> 1.8`)
- Erlang (`~> 21.3`)

:tada:

Closes #13

### DIFF
--- a/BOILERPLATE_README.md
+++ b/BOILERPLATE_README.md
@@ -22,11 +22,11 @@
 
 ## 🚧 Dependencies
 
-- Node.js (`10.14.2`)
-- NPM (`6.4.1`)
-- Elixir (`1.8.1`)
-- Erlang (`21.3.3`)
-- PostgreSQL (`~10.0`)
+- Node.js (`~> 10.14`)
+- NPM (`~> 6.9`)
+- Elixir (`~> 1.8`)
+- Erlang (`~> 21.3`)
+- PostgreSQL (`~> 10.0`)
 
 ## 🏎 Kickstart
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ before_install:
 before_script:
   - nvm install 10.14.2
   - nvm use 10.14.2
-  - npm install -g npm@6.4.1
+  - npm install -g npm@6.9.0
   - npm --prefix assets ci
   - mix compile
   - mix ecto.create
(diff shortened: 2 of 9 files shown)


BOILERPLATE_README.md
```diff
@@ -22,11 +22,11 @@
 
 ## 🚧 Dependencies
 
-- Node.js (`10.14.2`)
-- NPM (`6.4.1`)
-- Elixir (`1.8.1`)
-- Erlang (`21.3.3`)
-- PostgreSQL (`~10.0`)
+- Node.js (`~> 10.14`)
+- NPM (`~> 6.9`)
```
gcauchon: Node 10 is bundled with `6.4.1`, why are we upgrading to `6.9.0`?

Furthermore, we should probably upgrade to the latest `10.15.3`.

mandarvaze: +1 for upgrading to latest `10.15.3`

remiprev: > Node 10 is bundled with 6.4.1, why are we upgrading to 6.9.0?

Because we can do it easily 😄 Is there a reason we should _not_ do?

> Furthermore, we should probably upgrade to the latest 10.15.3.

Unfortunately, the latest `nodejs` and `npm` version on our Alpine version is `10.14.2-r0` (even if `10.15.3-r0` [seems](https://pkgs.alpinelinux.org/package/edge/main/x86/npm) to be available; I don’t know Alpine enough to upgrade to it.)

gcauchon: `10.15.3-r0` only is available on the `edge` branch, but `elixir:1.8.1-alpine` is based on Alpine `3.9`

https://pkgs.alpinelinux.org/package/v3.9/main/x86/nodejs

PR #18, opened 2019-04-08
Downgrade NPM version requirement to 6.4

remiprev:
Like @gcauchon suggested in https://github.com/mirego/elixir-boilerplate/pull/15, we enforced npm 6.9.0 for no valid reason. Let’s use the same NPM version as the one provided by Node 10.14.

### DIFF
--- a/assets/package.json
+++ b/assets/package.json
@@ -1,14 +1,18 @@
 {
-  "name": "",
+  "name": "elixir-boilerplate",
+  "version": "0.0.1",
   "private": true,
+  "engine-strict": true,
   "engines": {
-    "node": "~> 10.14",
-    "npm": "~> 6.9"
+    "node": "~10.14",
+    "npm": "~6.4"
   },
   "scripts": {
     "deploy": "npm run enforce-engine-versions && npx webpack --mode production",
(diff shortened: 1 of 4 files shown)


assets/package.json
```diff
@@ -3,12 +3,14 @@
...
-    "npm": "~> 6.9"
+    "npm": "~> 6.4"
   },
   "scripts": {
     "deploy": "npm run enforce-engine-versions && npx webpack --mode production",
     "watch": "npm run enforce-engine-versions && npx webpack --mode development --watch-stdin",
-    "enforce-engine-versions": "./scripts/enforce-engine-versions.js"
+    "enforce-engine-versions": "./scripts/enforce-engine-versions.js",
+    "preinstall": "npm run enforce-engine-versions",
+    "postinstall": "npm run enforce-engine-versions"
```
gcauchon: Out of curiosity, why are we adding two hooks in the `install` lifecycle?

remiprev: Removing `preinstall` was a typo I made in #15 — it shouldn’t have been removed. I should have added `postinstall` as well (because `semver` might not be installed in `preinstall`).

https://github.com/mirego/elixir-boilerplate/pull/15/files#diff-84a003e0defbcf4bc4b690d3004222b8L11

PR #21, opened 2019-04-10
Add “make push” target

remiprev:
Taken from [`react-boilerplate`](https://github.com/mirego/react-boilerplate/blob/64475d68443f00c0ca59ba7f3eeb3fa6182054f3/Makefile#L53-L55).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ APP_NAME = `grep 'app:' mix.exs | sed -e 's/\[//g' -e 's/ //g' -e 's/app://' -e 
 APP_VERSION = `grep 'version:' mix.exs | cut -d '"' -f2`
 GIT_REVISION = `git rev-parse HEAD`
 DOCKER_IMAGE_TAG ?= latest
+DOCKER_REGISTRY ?=
 
 # Introspection targets
 # ---------------------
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -53,6 +57,10 @@ dependencies-npm:
 build: ## Build the Docker image for the OTP release
 	docker build --build-arg APP_NAME=$(APP_NAME) --build-arg APP_VERSION=$(APP_VERSION) --rm --tag $(APP_NAME):$(DOCKER_IMAGE_TAG) .
 
+.PHONY: push
+push: ## Push the Docker image
+	docker push $(DOCKER_REGISTRY)/$(APP_NAME):$(DOCKER_IMAGE_TAG)
```
gcauchon: In order to push to `DOCKER_REGISTRY`, I think you first need to tag the image with `$(DOCKER_REGISTRY)/$(APP_NAME):$(DOCKER_IMAGE_TAG)`

remiprev: You’re right! For example if I want to push it as `mirego/elixir_boilerplate`, I have to : 

```
$ docker tag elixir_boilerplate:latest mirego/elixir_boilerplate:latest
$ DOCKER_REGISTRY=mirego make push
```

What do you suggest we do? Should `DOCKER_REGISTRY=mirego make push` tag **and** push the image?

remiprev: Or should we change the `build` target to this?

```bash
docker build --build-arg APP_NAME=$(APP_NAME) --build-arg APP_VERSION=$(APP_VERSION) --rm --tag $(DOCKER_REGISTRY)/$(APP_NAME):$(DOCKER_IMAGE_TAG) .
```

gcauchon: I think building an image and and pushing to a repo are 2 separate concerns! One might want to build the image locally to test the OTP release, but the CI process publish the images to the repo…

Changing the `push` target to do both `tag` and `push` with the `DOCKER_REGISTRY` url would make sense _imho_!

remiprev: Check it again! 😄

PR #22, opened 2019-04-11
Replace the HealthController by a plug that runs before middleware triggering redirects

pascalhamel:
When running the app in environments such as AWS Elastic Beanstalk, we need a health endpoint that can be called in HTTP, without enforcing the canonical host. In order to achieve that, we transform the `HealthController` in a plug that comes before middleware such as `:canonical_host`, `:force_ssl` and `:basic_auth`.


lib/elixir_boilerplate_web/health/plug.ex
```diff
@@ -0,0 +1,14 @@
+defmodule ElixirBoilerplateWeb.Health.Plug do
+  use Plug.Builder
+
+  import ElixirBoilerplate.Gettext
+
+  def call(%{request_path: "/health"} = conn, _) do
+    conn
+    |> put_resp_header("content-type", "text/html")
```
simonprev: It needs the HTML content type to succeed?

remiprev: I think we should actually enforces the practice of returning a serialized JSON object instead of HTML (or plain text content), what do you think?

gcauchon: I don't think the payload format actually matters since the load balancer only validates the `200` response.

But to continue along the previous comments, we probably should expose _metadata_ instead of "ok" and not rely on `dgettext/2` since this will never have to be translated.

I think returning the version number might be handy down the road for operational purposes… Something along those lines:

```elixir
def call(%{request_path: "/health"} = conn, _) do
  conn
  |> json(%{status: "ok", version: Application.spec(:elixir_boilerplate, :vsn)})
  |> halt()
```

remiprev: We won’t be able to use `json/2` since it’s we’re in a simple Plug and not a Phoenix controller action but we get the idea 😁

pascalhamel:
> This file should be deleted: https://github.com/mirego/elixir-boilerplate/blob/master/priv/gettext/en/LC_MESSAGES/health.po

Il appears it is being used in the `HomeController`. Should we change that or simply keep the file?

simonprev:
> Il appears it is being used in the HomeController

Oh right, leave it there, we should not change the home template 👍

PR #23, opened 2019-04-11
Actually compile assets in Docker build

remiprev:
We have to compile assets before running `mix phx.digest` in the Docker build process 😬


remiprev:
@simonprev I added actual CSS styles in `assets/css/app.scss` to make sure that we have a working assets setup in the boilerplate: https://github.com/mirego/elixir-boilerplate/pull/23/commits/04fe31cafc756536f9099ab1403cf3371bed0b72 😄

PR #26, opened 2019-04-11
Refactor errors view and fix health version

simonprev:
We were not using `use Sentry.Phoenix.Endpoint` as mentioned in `sentry-elixir` documentation. I also upgraded the sentry package. Since it is dependant of many packages, phoenix, hackney… were also updated (only minor version bumps).

I also sneaked a `Health.Plug` change 😄 

Here is the before/after:
image


lib/elixir_boilerplate_web/health/plug.ex
```diff
@@ -2,9 +2,14 @@ defmodule ElixirBoilerplateWeb.Health.Plug do
...
   def call(%{request_path: "/health"} = conn, _) do
+    version =
+      :elixir_boilerplate
+      |> Application.spec(:vsn)
+      |> String.Chars.to_string()
+
     conn
     |> put_resp_header("content-type", "application/json")
-    |> send_resp(200, Jason.encode!(%{status: "ok", version: Application.spec(:elixir_boilerplate, :vsn)}))
+    |> send_resp(200, Jason.encode!(%{status: "ok", version: version}))
```
gcauchon: Your right, running `Application.spec(:elixir_boilerplate, :vsn)` in `iex` was "misleading" since it automatically converts to string, which `Jason.encode/1` does not!

simonprev: > since it automatically converts to string

It renders the list of chars as `'1.0.0'` so it can be misleading yes but `iex` does not convert output to string (like `"1.0.0"`) 😄

gcauchon: This is what I meant, but I failed to explain it right 😬

PR #28, opened 2019-04-16
Upgrade to eslint 5

remiprev:
We now use the latest `eslint` major version channel.

~But we now also have to pass explicit sub-directories to lint because of [this bug](https://github.com/eslint/eslint/issues/9227) which prevents us from linting `assets` while ignoring `assets/node_modules` 🤷‍♂️~

Edit: We can now only lint the `assets` directory! We just needed to add the `assets` prefix in front of `assets/.eslintignore` entries.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,7 +82,7 @@ lint-credo:
 
 .PHONY: lint-eslint
 lint-eslint:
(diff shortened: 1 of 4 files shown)


gcauchon:
Sorry @charlesdemers I clicked _re-request review_ by mistake 😳

PR #29, opened 2019-04-16
Expose application version in config

remiprev:
We now have a way to expose the current application version (either fetched from `Mixfile` or from the OTP release).

### DIFF
--- a/rel/config/config.exs
+++ b/rel/config/config.exs
@@ -51,7 +51,7 @@ defmodule Utils do
       |> Application.spec(:vsn)
       |> case do
         nil -> Mix.Project.config()[:version]
-        version -> List.to_string(version)
+        version -> to_string(version)
       end
     end
   end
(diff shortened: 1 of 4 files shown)


rel/config/config.exs
```diff
@@ -64,6 +64,7 @@ port = Utils.Environment.get("PORT")
 
 # General application configuration
 config :elixir_boilerplate,
+  version: Utils.Version.get(),
```
gcauchon: Should we use this in the Healthcheck `plug` instead of `Application.spec(:elixir_boilerplate, :vsn) |> tostring()`?

remiprev: Yep, good catch! I’ll change it in this branch.

remiprev:
@simonprev @gcauchon I made two changes since your last reviews:

* We now use `String.Chars.to_string/1` instead of `List.to_chars/1` to convert the version to a string
* We now use the exposed config version in the health plug response

rel/config/config.exs
```diff
@@ -51,7 +51,7 @@ defmodule Utils do
       |> Application.spec(:vsn)
       |> case do
         nil -> Mix.Project.config()[:version]
-        version -> List.to_string(version)
+        version -> String.Chars.to_string(version)
```
gcauchon: I think we can even use the `to_string/1`macro directly!

https://hexdocs.pm/elixir/Kernel.html#to_string/1

PR #33, opened 2019-04-23
Ditch config/prod.exs file

remiprev:
The `config/prod.exs` file was never used (edit: _at runtime_!) in production (since we deploy an OTP release).

I emptied and left the file there though, to prevent developers from adding configuration in it and wondering why it’s not working in production. I also left a note explaining the situation 😄


gcauchon:
I am not sure if the centralized approach is right, especially since we need to add an `if` statement to branch config. The "standard" setup is already handled in [`config/config.exs`](https://github.com/mirego/elixir-boilerplate/blob/master/config/config.exs#L8) :

```elixir
import_config "#{Mix.env()}.exs"
```

remiprev:
> The "standard" setup is already handled in config/config.exs

No, `config/config.exs` is not used in the OTP release, only `rel/config/config.exs` [is used](https://github.com/mirego/elixir-boilerplate/blob/master/rel/config.exs#L35-L45). Isn’t it?

gcauchon:
From what I can remember, when `server: true` and `root: "."` where missing from `config/prod.exs`, the Endpoint application was not started in the OTP release. Adding the missing lines made it work, which made me think the `config/*.exs` configurations were handled by the `Mix` config provider…

remiprev:
Ah! `config/prod.exs` is only evaluated at _build time_, not at _run time_ (when the release is ran).

What do you think? I still think we should enforce _not using it_, since it could create confusion (as it already did in one of our project 😄).

PR #35, opened 2019-04-25
Proposition : environment/configuration improvements

gcauchon:
The original setup had a few flaws:

1. Move back _compile time_ configuration to `config/*.exs` to keep only the _runtime_ configuration in `rel/config/release.exs`. This better separate concern and also gets us in a better position  for the mix-based release flow that is going to land in Elixir `1.9`!

1. We were overusing pattern matching in the `Utils.Environment` _helper_ module. And the `Utils.` nesting does not serve any significant meaning.

1. Running the application with the vanilla `docker-compose` and `.env.dev` caused an endpoint error:
    
    ``` 
    Request: GET /
    ** (exit) an exception was raised:
        ** (ArgumentError) cookie store expects conn.secret_key_base to be at least 64 bytes
        …
    ```
    ⮑ Since this was this was "by design", add a step in `BOILERPLATE_README.md` to mention secrets needs to be generated with `mix phx.gen.secret`.


remiprev:
> Running the application with the vanilla docker-compose and .env.dev caused an endpoint error:

Yes, this is by design. We don’t want to hardcode secrets in `.env.dev`. We want developers using this boilerplate project to have to generate them.

config/prod.exs
```diff
@@ -1,8 +1,13 @@
...
-# block in `rel/config/config.exs` since we use an OTP release to deploy and
-# run the application in production.
-#
-# Configuration specified in this file would only be picked up at runtime when
-# running `mix phx.server` in production, which we never do.
+use Mix.Config
+
+# Configure endpoint
+config :elixir_boilerplate, ElixirBoilerplateWeb.Endpoint,
+  cache_static_manifest: "priv/static/cache_manifest.json",
```
remiprev: This reintroduces the bug fixed by #33 — The `cache_static_manifest` configuration on `ElixirBoilerplateWeb.Endpoint` must be set at _runtime_ not at _compile time_.

config/config.exs
```diff
@@ -1,8 +1,21 @@
...
+# Configure endpoint
+config :elixir_boilerplate, ElixirBoilerplateWeb.Endpoint,
+  pubsub: [name: ElixirBoilerplate.PubSub, adapter: Phoenix.PubSub.PG2],
+  render_errors: [view: ElixirBoilerplateWeb.Errors.View, accepts: ~w(html json)]
+
+# Configure Gettext
+config :elixir_boilerplate, ElixirBoilerplate.Gettext, default_locale: "en"
+
+# Import runtime configuration
+import_config "../rel/config/release.exs"
```
remiprev: I think we should be a bit more explicit and name the file `runtime.exs` since:

* The file is used to store runtime configuration
* The file is not only used to store configuration that will be used in the release — it will also be used when running `mix phx.server` in all environments

.env.dev
```diff
@@ -15,9 +15,9 @@ CANONICAL_HOST=localhost
 DEBUG_ERRORS=true
 FORCE_SSL=false
 PORT=4000
-SECRET_KEY_BASE= # Generate with `mix phx.gen.secret`
-SESSION_KEY= # Generate with `mix phx.gen.secret`
-SIGNING_SALT= # Generate with `mix phx.gen.secret`
+SECRET_KEY_BASE=__Generate_a_64_bytes_secret_with_phoenix_generator:_mix phx.gen.secret__
+SESSION_KEY=__Generate_a_session_key_with_mix phx.gen.secret__
+SIGNING_SALT=__Generate_a_signing_salt_with_mix phx.gen.secret__
```
remiprev: > Running the application with the vanilla docker-compose and .env.dev caused an endpoint error:

Yes, this is by design. We don’t want to hardcode secrets in .env.dev. We want developers using this boilerplate project to have to generate them.

gcauchon: I will add instruction in the `Readme` then to explain this mandatory step for both development and testing the OTP.

gcauchon: I will add instruction in the `Readme` then to explain this mandatory step for both development and testing the OTP.

simonprev: Since you make modification, can we remove the `mix phx.gen.secret` on `SESSION_KEY`? The session key is not meant to be stored as a secret key. It should be set as `elixir-boilerplate`. It will then be renamed to the app name on the setup script.

rel/config/release.exs
```diff
@@ -0,0 +1,87 @@
...
+    ]
+end
+
+# Configure Sentry
+config :sentry,
+  dsn: Environment.get("SENTRY_DSN"),
+  environment_name: Environment.get("SENTRY_ENVIRONMENT_NAME"),
+  included_environments: [:prod],
+  root_source_code_path: File.cwd!(),
+  release: Environment.version()
```
remiprev: We could split it in both files

### `config/config.exs`

```elixir
# Configure Sentry
config :sentry,
  included_environments: [:prod],
  root_source_code_path: File.cwd!(),
  release: Environment.version()
```

### `rel/config/runtime.exs`

```elixir
# Configure Sentry
config :sentry,
  dsn: Environment.get("SENTRY_DSN"),
  environment_name: Environment.get("SENTRY_ENVIRONMENT_NAME")
```

config/prod.exs
gcauchon: After further investigation, `mix release` correctly generate a `sys.config` file in the OTP release with configuration from `config/*.exs`! Runtime configuration get merged in when the `Mix.Releases.Config.Providers.Elixir` runs at boot time to parse the `rel/config/runtime.exs` file that gest overlayed in the bundle.

config/config.exs
```diff
@@ -1,8 +1,21 @@
...
+
+# Import environment configuration
 import_config "#{Mix.env()}.exs"
```
remiprev: We probably should have the environment-specific imported _before_ the runtime configuration, right?

README.md
```diff
@@ -27,6 +27,17 @@ This boilerplate comes with batteries included, you’ll find:
...
+
+Before starting the server with `mix phx.server` or testing the OTP/Docker setup with `make dev-start`, generate secret values for the following keys in [`.env.dev`](./.env.dev):
+    * `SECRET_KEY_BASE`
+    * `SIGNING_SALT`
+
+```shell
+> mix phx.gen.secret
+Eif/HWo9DB0A8h+tYYJDZOZUKXVTzLnHa7faTFySn6eIqocYqbcE8Hs9C9qp98Po
+```
+
```
gcauchon: @remiprev Do you think this be enough?

README.md
```diff
@@ -22,10 +22,22 @@ This boilerplate comes with batteries included, you’ll find:
...
 
 1. Clone this project (`git clone https://github.com/mirego/elixir-boilerplate.git`)
-2. Delete the internal Git directory (`rm -rf .git`)
-3. Run the boilerplate setup script (`./boilerplate-setup.sh YourProjectName`)
-4. Create a new Git repository (`git init`)
-5. Create the initial Git commit (`git commit -a -m "Initial commit"`)
+1. Delete the internal Git directory (`rm -rf .git`)
+1. Run the boilerplate setup script (`./boilerplate-setup.sh YourProjectName`)
+1. Create a new Git repository (`git init`)
+1. Create the initial Git commit (`git commit -a -m "Initial commit"`)
```
remiprev: We shouldn’t touch these numbers.

README.md
remiprev: This should be added to `BOILERPLATE_README.md`, not `README.md` 😄

I would also try to avoid listing specific commands and specific variables. We already have [this line](https://github.com/mirego/elixir-boilerplate/blob/master/BOILERPLATE_README.md#initial-setup): 

> Create both .env.dev.local and .env.test.local from empty values in .env.dev and .env.test

We should just add a small note mentioning the need to populate missing variables (instructions are already in the `.env.dev` file).

BOILERPLATE_README.md
```diff
@@ -40,9 +40,10 @@ When running `mix` or `make` commands, it is important that these variables are
 
 1. Create both `.env.dev.local` and `.env.test.local` from empty values in [`.env.dev`](./.env.dev) and [`.env.test`](./.env.test)
 2. Install Mix and NPM dependencies with `make dependencies`
-3. Create and migrate the database with `mix ecto.setup`
-4. Compile the application with `make compile`
-5. Start the Phoenix server with `iex -S mix phx.server` with environment variables from `.env.dev` and `.env.dev.local`
+3. Generate values for mandatory secrets in [`.env.dev`](./.env.dev) with `mix phx.gen.secret`
```
gcauchon: @remiprev It makes more sense here!

.travis.yml
```diff
@@ -26,6 +26,8 @@ env:
     - MIX_ENV: test
     - DATABASE_URL: postgres://elixir_boilerplate:password@localhost:5433/elixir_boilerplate_test
     - PGPORT: 5433
+    - SECRET_KEY_BASE: yv8eF+KYEbap8J8I9h1bLtWliN/tF6uFa5ba+/pxrTmDYf9kCUMSpb1R2RYTKQ8V 
+    - SIGNING_SALT: 2ly5tiMCHPachCPOIQjMcrk6PxRaXA25QFpaspnuumMyJZgAyhH8O1d59FluvwyS
```
remiprev: Oh… this is a drawback from having `config/test.exs` (which sets these values) **before** `rel/config/runtime.exs` 😬 

I think we have to find a way to avoid doing this.

rel/config/runtime.exs
```diff
@@ -0,0 +1,85 @@
...
+# Configure application
+force_ssl = Environment.get_boolean("FORCE_SSL")
+scheme = if force_ssl, do: "https", else: "http"
+host = Environment.get("CANONICAL_HOST")
+port = Environment.get("PORT")
+
+config :elixir_boilerplate,
+  canonical_host: host,
+  force_ssl: force_ssl,
+  version: Environment.version()
```
remiprev: Since this doesn’t come from the runtime environment, we should put this in `config/config.exs` shouldn’t we?

And since the version in compiled at _build time_, we can safely use `Mix.Project.config()[:version]` 🎉  because `Application.spec(:elixir_boilerplatecase, :vsn)` is only available at runtime.

```elixir
# Configure application
config :elixir_boilerplate,
  ecto_repos: [ElixirBoilerplate.Repo],
  version: Mix.Project.config()[:version]
```

gcauchon: Sure! No need for `Environment.version/0` anymore 👍

rel/config/runtime.exs
```diff
@@ -0,0 +1,76 @@
...
+
+  def get_boolean(key) do
+    case get(key) do
+      "true" -> true
+      "1" -> true
+      _ -> false
+    end
+  end
+
+  def get_integer(key, default \\ 0) do
```
remiprev: Other commands don’t have a default — I think we should return `nil` if the key doesn’t exist.

gcauchon: Consistency! - @garno ™️

BOILERPLATE_README.md
```diff
@@ -40,9 +40,10 @@ When running `mix` or `make` commands, it is important that these variables are
...
+4. Create and migrate the database with `mix ecto.setup`
+5. Compile the application with `make compile`
```
remiprev: This should be `mix compile`. Should we include this at all? Running `iex -S mix phx.server` will compile it anyway.

PR #36, opened 2019-04-25
Add Corsica default configuration

### LINKED ISSUE #10: Add corsica
Since we use [Corsica](https://hex.pm/packages/corsica) in the majority of our project, we could use it in this boilerplate like we do `plug_canonical_host` or `basic_auth` — only _enable_ it if a `CORS_ALLOWED_ORIGINS` environment variable is present.

garno:
Since we mostly create API for SPA, we add the default configuration needed for CORS with Corsica.

Closes #10

### DIFF
--- a/lib/elixir_boilerplate_web/endpoint.ex
+++ b/lib/elixir_boilerplate_web/endpoint.ex
@@ -9,9 +9,9 @@ defmodule ElixirBoilerplateWeb.Endpoint do
   )
 
   plug(ElixirBoilerplateWeb.Health.Plug)
-  plug(:cors)
   plug(:canonical_host)
   plug(:force_ssl)
+  plug(:cors)
   plug(:basic_auth)
   plug(:session)
 
--- a/lib/elixir_boilerplate_web/endpoint.ex
+++ b/lib/elixir_boilerplate_web/endpoint.ex
@@ -9,6 +9,7 @@ defmodule ElixirBoilerplateWeb.Endpoint do
   )
 
   plug(ElixirBoilerplateWeb.Health.Plug)
+  plug(:cors)
   plug(:canonical_host)
   plug(:force_ssl)
   plug(:basic_auth)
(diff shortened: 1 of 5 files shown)


rel/config/config.exs
```diff
@@ -111,6 +111,11 @@ if Utils.Environment.exists?("BASIC_AUTH_USERNAME") do
     ]
 end
 
+# Configure CORS
+config :tv5, :corsica,
+  origins: System.get_env("CORS_ALLOWED_ORIGINS"),
```
remiprev: What is `CORS_ALLOWED_ORIGINS` exactly? A string of comma-separated origins? Eg. `"foo.com,bar.com"`?

rel/config/config.exs
```diff
@@ -111,6 +111,11 @@ if Utils.Environment.exists?("BASIC_AUTH_USERNAME") do
     ]
 end
 
+# Configure CORS
+config :elixir_boilerplate, :corsica,
+  origins: System.get_env("CORS_ALLOWED_ORIGINS"),
```
remiprev: What is `CORS_ALLOWED_ORIGINS` exactly? A comma-separated list of origins? Eg. `"foo.com,bar.com"`?

garno: Good question, for now it’s a single string but we should probablement split it if we want to support a comma-separated list of origins as you mentioned.

remiprev: I just looked at the [Origins](https://hexdocs.pm/corsica/Corsica.html#module-origins) section and for multiple origins it’s:

```elixir
plug Corsica, origins: ["http://foo.com", "http://bar.com"]
```

This is not possible with the current setup, is it?

remiprev:
```suggestion
  origins: String.split(System.get_env("CORS_ALLOWED_ORIGINS"), ","),
```

And I’m guessing `["*"]` will work 👍

remiprev: We should probably refrain from using `System.get_env` like the rest of the file. We could have a `Utils.Environment.get_list("SENTRY_DSN")` which would do the heavy lifting for us 😄

garno: Yes, I’m gonna wait for PR #35 to be merged and will complete the setup afterward.

remiprev:
@garno Just need to rebase when you get the chance and I think it’s ready to merge!

lib/elixir_boilerplate_web/endpoint.ex
```diff
@@ -9,6 +9,7 @@ defmodule ElixirBoilerplateWeb.Endpoint do
   )
 
   plug(ElixirBoilerplateWeb.Health.Plug)
+  plug(:cors)
```
remiprev: I think `cors` should be just after `canonical_host` and `force_ssl`. I don’t think we want to respond to requests without going through these two plugs first.

PR #39, opened 2019-05-08
Relax Node.js and NPM version requirements

### LINKED ISSUE #13: Strict version requirements

remiprev:
We now enforce less strict engine version requirements for Node.js and NPM.

I don’t think it’s a high-risk change since there should never be any breaking changes between minor Node.js and NPM releases.

Maybe this is the first step in moving _all_ version requirements to major versions (eg. Elixir `1.*` and Erlang `21.*`)  🤷‍♂ 

I would love to hear @charlesdemers and @simonprev take on this 👋 

Closes #13

### DIFF
--- a/BOILERPLATE_README.md
+++ b/BOILERPLATE_README.md
@@ -22,8 +22,8 @@
 
 ## 🚧 Dependencies
 
-- Node.js (`~> 10.14`)
-- NPM (`~> 6.4`)
+- Node.js (`>= 10.14, < 11.0`)
+- NPM (`>= 6.4, < 7.0`)
 - Elixir (`~> 1.8`)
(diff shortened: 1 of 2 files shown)


remiprev:
> So: Erlang 21.* and Elixir 1.8.* 👍

Alright, I’ll do them next :grinning:

Issue #38, opened 2019-05-04
query : http entry missing from config/dev.exs


mandarvaze:
(I'm new to elixir/phoenix development - so please pardon my ignorance)

As the title says http entry is missing from `dev.exs`, without it I get `(RuntimeError) server can't start because :port in http config is nil, please use a valid port number` when I run `iex -S phx.server`

When I added the following line to `dev.exs` the error went away

`http: [port: System.get_env("PORT") || 4000`

(Found the above [here](https://github.com/phoenixframework/phoenix/issues/1355#issuecomment-157218229))

On a related note : Is there a reason `config/dev.exs` and `config/test.exs` so different ?
`test.exs` has secret and signing salt hard coded, while `dev` (is supposed to , I think) get it from `.env.dev.local`

remiprev:
Hi Mandar!

Are you sure `.env.dev` is evaluated by the process that starts your server? The default value [is present in `.env.dev`](https://github.com/mirego/elixir-boilerplate/blob/master/.env.dev#L17).

> On a related note : Is there a reason config/dev.exs and config/test.exs so different ?

Yes, there is 😄 It’s to be able to run `mix test` without providing _all the environment variables_ the application requires. Every development setup is different while we try to make sure that the testing setup is the same (and straightforward to run) whether it’s on a developer machine or on a CI environment.

Hope that helps!

mandarvaze:
Yeah. the value is **defined** in `.env.dev`, but is not referenced/used in `dev.exs`

remiprev:
Yes, but here’s the sequence of config files loaded by the application:

1. `config/config.exs`
2. `rel/config/runtime.exs`
3. `config/<env>.exs`

The value is used in [`rel/config/runtime.exs`](https://github.com/mirego/elixir-boilerplate/blob/master/rel/config/runtime.exs#L37) because it comes from an environment variable.

mandarvaze:
Does `export` work with `nv` ? ( I guess not, but I have not used `nv`)

Keeping **two** copies of the `.env` files - one for `source` (with export) and one for `nv` (without export) is not worth it. Just choose one (whichever you prefer)

Personally I prefer `source` (cause it comes built-in with the shell, while `nv` is external)

remiprev:
The key instruction is:

> When running `mix` or `make` commands, it is important that these variables are present in the environment.

[...]

PR #41, opened 2019-05-10
Updated documentation for #38

mandarvaze:
Updated instructions related to using `source` for setting up environment variables


remiprev:
Hi Mandar,

I’m going to close this since I updated the documentation to avoid using `source`: https://github.com/mirego/elixir-boilerplate/commit/0f25517213b356209071e676161672882b0791b6

I think this will avoid any confusion by just stating that the variables needs to be in the environment.

Thank you!

PR #42, opened 2019-05-15
Move toward the future mix-based release configuration

gcauchon:
Releases will be built-in `mix` in Elixir 1.9 so we are moving toward the configuration setup documented to simplify the future migration from Distillery…

⮑ https://github.com/elixir-lang/elixir/pull/8677

## tl;dr

https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/release.ex#L17-L31

> A release can be configured in your `mix.exs` file under the `:releases` key inside `def project`

https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/release.ex#L477-L480

> To enable runtime configuration in your release, all you need to do is to create a file named `config/releases.exs`


.credo.exs
```diff
@@ -54,7 +54,7 @@
         {Credo.Check.Warning.UnusedTupleOperation},
         {Credo.Check.Warning.OperationWithConstantResult},
         {CredoEnvvar.Check.Warning.EnvironmentVariablesAtCompileTime},
-        {CredoFilenameConsistency.Check.Consistency.FilenameConsistency, excluded_paths: ["test/support", "priv", "rel", "mix.exs"]}
+        {CredoFilenameConsistency.Check.Consistency.FilenameConsistency, excluded_paths: ["mix.exs", "config/releases.exs", "priv", "rel", "test/support"]}
```
remiprev: I think we’ll be able to remove `"rel"` from the list.

gcauchon: I also added the whole `config` folder, not only `config/releases.exs` from the exclusions since modules created in config script files are utilities, not "runtime" critical modules. Thoughts?

PR #45, opened 2019-05-16
Add support for Erlang version with extra digits

remiprev:
Erlang versions can be:

* `21.1`
* `21.1.2`
* `21.1.2.3`

Only the second one is a valid SemVer version. We already supported `21.1` (by remapping it to `21.1.0`) and now we support `21.1.2.3` too (by remapping it to `21.1.2`).

### DIFF
--- a/lib/mix/tasks/erlang.check_version.ex
+++ b/lib/mix/tasks/erlang.check_version.ex
@@ -30,23 +30,23 @@ defmodule Mix.Tasks.Erlang.CheckVersion do
     |> Path.join()
     |> File.read()
     |> case do
-      {:ok, version} -> sanitize_version(String.trim(version))
+      {:ok, version} -> sanitize_version(version)
       _ -> nil
     end
   end
 
+  # OTP versions are not always compatible with Semantic Versioning. For
+  # example, `21.3` and `21.3.0.1` are valid OTP versions, but not valid
+  # “semver” ones. We need to set them both to `21.3.0` to match it against our
+  # expected version requirement.
   defp sanitize_version(version) do
-    ~r/^\d+\.\d+$/
-    |> Regex.run(version)
+    version
+    |> String.trim()
+    |> String.split(".")
     |> case do
-      [version] ->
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/erlang.check_version.ex
```diff
@@ -30,12 +30,18 @@ defmodule Mix.Tasks.Erlang.CheckVersion do
     |> Path.join()
     |> File.read()
     |> case do
-      {:ok, version} -> sanitize_version(String.trim(version))
-      _ -> nil
+      {:ok, version} ->
+        version
+        |> String.trim()
+        |> sanitize_version_without_patch()
+        |> sanitize_version_with_extra_digits()
```
simonprev: Can’t we just do "21.1.2" split with ".", Pad with "0" (or remove elements to have 3 elements) and join with "."?

21.1 would give 21.1.0
21.1.2.3.4 would give 21.1.2

Do not we really need regex for that?

gcauchon: I like @simonprev proposition!

```elixir
version
|> String.trim()
|> String.split(".")
|> Enum.concat(~w(0 0)) 
|> Enum.slice(0, 3) 
|> Enum.join(".")
```

lib/mix/tasks/erlang.check_version.ex
```diff
@@ -30,23 +30,23 @@ defmodule Mix.Tasks.Erlang.CheckVersion do
...
-        # OTP versions are not always compatible with Semantic Versioning. For example,
-        # `21.3` is a valid OTP version, but is not a valid “semver” one. We need to set
-        # it to `21.3.0` to match it against our expected version requirement.
-        "#{version}.0"
-
-      nil ->
-        version
+      [major] -> "#{major}.0.0"
+      [major, minor] -> "#{major}.#{minor}.0"
+      [major, minor, patch | _] -> "#{major}.#{minor}.#{patch}"
```
gcauchon: Ohhh, I like the use of pattern matching even more!

thermech:
👀

PR #46, opened 2019-05-21
Add user-password placeholders in the DATABASE_URL

mandarvaze:
This helps prevent `KeyError exception: key :password not found.` Without the username/password we get the following error when the server is started :

```
[error] GenServer #PID<0.558.0> terminating
** (RuntimeError) connect raised KeyError exception: key :password not found.The exception details are hidden, as they may contain sensitive data such as database credentials. You may set :show_sensitive_data_on_connection_error to true if you wish to see all of the details
    (elixir) lib/keyword.ex:389: Keyword.fetch!/2
    (postgrex) lib/postgrex/protocol.ex:719: Postgrex.Protocol.auth_md5/4
    (postgrex) lib/postgrex/protocol.ex:576: Postgrex.Protocol.handshake/2
    (db_connection) lib/db_connection/connection.ex:66: DBConnection.Connection.connect/2
    (connection) lib/connection.ex:622: Connection.enter_connect/5
    (stdlib) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
Last message: nil
State: Postgrex.Protocol
```

### DIFF
--- a/.env.dev
+++ b/.env.dev
@@ -20,7 +20,7 @@ SESSION_KEY=elixir_boilerplate
 SIGNING_SALT= # Generate salt with `mix phx.gen.secret`
 
 # Database configuration
-DATABASE_URL=postgres://localhost/elixir_boilerplate_dev
+DATABASE_URL=postgres://username:password@localhost/elixir_boilerplate_dev
 DATABASE_POOL_SIZE=20
 
 # Basic Authentication


remiprev:
Hi Mandar,

You’ll have to provide more details like the actual `DATABASE_URL` used. This does not seem like a bug with `elixir-boilerplate`.

Thank you!

mandarvaze:
The default DATABASE_URL from `.env.dev` looks like :

`DATABASE_URL=postgres://localhost/project_dev`

When I added `username:password@` before the localhost, error went away.

While it may not be a "bug" - IMO having a placeholder in the default `.env.dev` will remind users to replace it with actual values of the username and password in their `.env.dev.local`

gcauchon:
Since the `docker-compose` setup includes `postgres:development` in the URL, I think it would make sense to preconfigure the `.env.dev` too.

https://github.com/mirego/elixir-boilerplate/blob/master/docker-compose.yml#L8

remiprev:
We use `postgres://localhost/elixir_boilerplate_dev` as the default value because when you install PostgreSQL, it runs by default with your username and without a password. So if you have

```
DATABASE_URL=postgres://localhost/elixir_boilerplate_dev
```

running `mix ecto.create` and `mix phx.server` should work out-of-the-box.

If we had `postgres://username:password@localhost/elixir_boilerplate_dev` as the default value, it may create confusion for new users as they don’t have a password.

I’ll add the username/password credentials bit as an example!

PR #53, opened 2019-05-29
Set max size to 40 lines for ABCSize credo check rule

romarickb:
By default, max_size for `Credo.Check.Refactor.ABCSize` is **30 lines**.
As it's a bit restrictive, we propose to increase it to **40 lines**


Boubalou:
I got scared by the title of the PR! But the implementation is all good :)

PR #54, opened 2019-05-29
Fix Corsica configuration to support "*" as an environment variable

gcauchon:
👀 Follow up to #36

Following the [Corsica](https://hexdocs.pm/corsica/Corsica) documentation, the `origins` option handle the "allow all" in a specific way to whitelist request from any website domain.

> The value `"*"` can also be used to match every origin and reply with `*` as the value of the `access-control-allow-origin` header. If `"*"` is used, it must be used as the only value of `:origins` (**that is, it can’t be used inside a list of accepted origins**).

With `CORS_ALLOWED_ORIGINS=*` set, here is the difference :

## Before

```shell
> curl --request OPTIONS --header "Origin: http://foo.bar" --header "Access-Control-Request-Method: POST" --verbose http://localhost:4000
…
> OPTIONS / HTTP/1.1
> Host: localhost:4000
> User-Agent: curl/7.54.0
> Accept: */*
> Origin: http://foo.bar
> Access-Control-Request-Method: POST
>
< HTTP/1.1 200 OK
< cache-control: max-age=0, private, must-revalidate
< content-length: 0
< date: Wed, 29 May 2019 18:43:27 GMT
< server: Cowboy
```

## After

```shell
> curl --request OPTIONS --header "Origin: http://foo.bar" --header "Access-Control-Request-Method: POST" --verbose http://localhost:4000
…
> OPTIONS / HTTP/1.1
> Host: localhost:4000
> User-Agent: curl/7.54.0
> Accept: */*
> Origin: http://foo.bar
> Access-Control-Request-Method: POST
>
< HTTP/1.1 200 OK
< access-control-allow-headers: 
< access-control-allow-methods: PUT, PATCH, DELETE
< access-control-allow-origin: *
< cache-control: max-age=0, private, must-revalidate
< content-length: 0
< date: Wed, 29 May 2019 18:41:29 GMT
< server: Cowboy
```

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -12,9 +12,9 @@ config :elixir_boilerplate, ElixirBoilerplateWeb.Endpoint,
   pubsub: [name: ElixirBoilerplate.PubSub, adapter: Phoenix.PubSub.PG2],
   render_errors: [view: ElixirBoilerplateWeb.Errors.View, accepts: ~w(html json)]
 
-config :elixir_boilerplate, ElixirBoilerplate.Gettext, default_locale: "en"
+config :elixir_boilerplate, Corsica, allow_headers: :all
 
-config :elixir_boilerplate, :corsica, allow_headers: :all
+config :elixir_boilerplate, ElixirBoilerplate.Gettext, default_locale: "en"
 
 config :sentry,
   included_environments: ~w(prod)a,
--- a/config/releases.exs
+++ b/config/releases.exs
@@ -22,10 +22,13 @@ defmodule Environment do
     end
   end
 
-  def get_list(key) do
-    case get(key) do
-      value when is_bitstring(value) -> String.split(value, ",")
-      _ -> []
+  def get_list_or_first_value(key) do
+    with value when is_bitstring(value) <- get(key),
+         [single_value] <- String.split(value, ",") do
+      single_value
+    else
+      values when is_list(values) -> values
+      _ -> nil
     end
   end
 
@@ -65,6 +68,8 @@ config :elixir_boilerplate, ElixirBoilerplateWeb.Endpoint,
   ],
   url: [scheme: scheme, host: host, port: port]
 
+config :elixir_boilerplate, Corsica, origins: Environment.get_list_or_first_value("CORS_ALLOWED_ORIGINS")
+
 if Environment.exists?("BASIC_AUTH_USERNAME") do
   config :elixir_boilerplate,
     basic_auth: [
@@ -72,8 +77,6 @@ if Environment.exists?("BASIC_AUTH_USERNAME") do
       password: Environment.get("BASIC_AUTH_PASSWORD")
     ]
 end
-
-config :elixir_boilerplate, :corsica, origins: Environment.get_list("CORS_ALLOWED_ORIGINS")
 
 config :sentry,
   dsn: Environment.get("SENTRY_DSN"),
--- a/lib/elixir_boilerplate_web/endpoint.ex
+++ b/lib/elixir_boilerplate_web/endpoint.ex
@@ -80,6 +80,12 @@ defmodule ElixirBoilerplateWeb.Endpoint do
     end
   end
 
+  defp cors(conn, _opts) do
+    opts = Corsica.init(Application.get_env(:elixir_boilerplate, Corsica))
+
+    Corsica.call(conn, opts)
+  end
+
   defp basic_auth(conn, _opts) do
     basic_auth_config = Application.get_env(:elixir_boilerplate, :basic_auth)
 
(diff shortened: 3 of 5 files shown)


config/config.exs
```diff
@@ -12,9 +12,9 @@ config :elixir_boilerplate, ElixirBoilerplateWeb.Endpoint,
   pubsub: [name: ElixirBoilerplate.PubSub, adapter: Phoenix.PubSub.PG2],
   render_errors: [view: ElixirBoilerplateWeb.Errors.View, accepts: ~w(html json)]
 
-config :elixir_boilerplate, ElixirBoilerplate.Gettext, default_locale: "en"
+config :elixir_boilerplate, Corsica, allow_headers: :all
 
-config :elixir_boilerplate, :corsica, allow_headers: :all
+config :elixir_boilerplate, ElixirBoilerplate.Gettext, default_locale: "en"
```
remiprev: Why do we have to change the order in `config/config.exs`?

config/releases.exs
```diff
@@ -73,8 +78,6 @@ if Environment.exists?("BASIC_AUTH_USERNAME") do
     ]
 end
 
-config :elixir_boilerplate, :corsica, origins: Environment.get_list("CORS_ALLOWED_ORIGINS")
```
remiprev: Same here? 🤔

config/config.exs
gcauchon: It's purely aesthetic, but it made no sense to me to see the GetText config between `Endpoint` and `Corsica`.

config/releases.exs
gcauchon: Same answer here, it make more sense to me to define the Corsica configuration right after Endpoint, before an optional blok to enable basic auth.

lib/elixir_boilerplate_web/endpoint.ex
```diff
@@ -80,6 +80,12 @@ defmodule ElixirBoilerplateWeb.Endpoint do
     end
   end
 
+  defp cors(conn, _opts) do
+    opts = Corsica.init(Application.get_env(:elixir_boilerplate, Corsica))
+
+    Corsica.call(conn, opts)
+  end
+
```
gcauchon: @remiprev As for the reordering here, I use the same order we have in the Plug lifecycle…

```elixir
plug(ElixirBoilerplateWeb.Health.Plug)
  plug(:canonical_host)
  plug(:force_ssl)
  plug(:cors)
  plug(:basic_auth)
  plug(:session)
```

mix.lock
```diff
@@ -1,47 +1,48 @@
...
+  "artificery": {:hex, :artificery, "0.4.2", "3ded6e29e13113af52811c72f414d1e88f711410cac1b619ab3a2666bbd7efd4", [:mix], [], "hexpm"},
+  "basic_auth": {:hex, :basic_auth, "2.2.4", "d8c748237870dd1df3bc5c0f1ab4f1fad6270c75472d7e62b19302ec59e92a79", [:mix], [{:plug, "~> 0.14 or ~> 1.0", [hex: :plug, repo: "hexpm", optional: false]}], "hexpm"},
   "bunt": {:hex, :bunt, "0.2.0", "951c6e801e8b1d2cbe58ebbd3e616a869061ddadcc4863d0a2182541acae9a38", [:mix], [], "hexpm"},
   "certifi": {:hex, :certifi, "2.5.1", "867ce347f7c7d78563450a18a6a28a8090331e77fa02380b4a21962a65d36ee5", [:rebar3], [{:parse_trans, "~>3.3", [hex: :parse_trans, repo: "hexpm", optional: false]}], "hexpm"},
   "connection": {:hex, :connection, "1.0.4", "a1cae72211f0eef17705aaededacac3eb30e6625b04a6117c1b2db6ace7d5976", [:mix], [], "hexpm"},
+  "corsica": {:hex, :corsica, "1.1.2", "5ad8b9dcbeeda4762d78a57c0c8c2f88e1eef8741508517c98cb79e0db1f107d", [:mix], [{:plug, "~> 1.0", [hex: :plug, repo: "hexpm", optional: false]}], "hexpm"},
```
remiprev: Corsica was _not_ in `mix.lock`? That’s weird. It’s OK to add it in this PR but I don’t think this PR should any other dependency requirements.

gcauchon: I ran `mix deps.update --all` to check the delta with Hex to realize Corsica was missing, but I forgot to rollback the other minor/patch changes… Should I?

remiprev: Yes I think so. If we want to upgrade other dependencies, we should do it in another PR 😄

gcauchon: It was already rollbacked when you replied 😎

PR #55, opened 2019-05-31
Add fully-functional Ecto error messages handling

remiprev:
This is the helper module we use in a few projects to handle Ecto error messages.

We can use it this way:

```eex
<%= if @changeset.action do %>
  <%= error_messages(@changeset) %>
<% end %>
```

to output something like:

```html
<ul>
  <li>email is invalid</li>
  <li>email is already taken</li>
  <li>role.type is not valid</li>
</ul>
```

I added basic tests but I’ll keep on adding more before merging because it doesn’t fully support “many” fields like `has_many` and `embeds_many` associations.

### DIFF
--- a/lib/elixir_boilerplate_web/errors/helpers.ex
+++ b/lib/elixir_boilerplate_web/errors/helpers.ex
@@ -1,43 +1,59 @@
 defmodule ElixirBoilerplateWeb.Errors.Helpers do
-  @moduledoc """
-  Conveniences for translating and building error messages.
-  """
-  use Phoenix.HTML
+  alias Ecto.Changeset
+  alias ElixirBoilerplateWeb.Errors.View
 
   @doc """
-  Generates tag for inlined form input errors.
+  Generates a human-readable block containing all errors in a changeset. Errors
+  are then localized using translations in the `ecto` domain.
+
+  For example, you could have an `ecto.po` file in the french locale:
+
+  ```
+  msgid ""
+  msgstr ""
+  "Language: fr"
+
+  msgid "can't be blank"
+  msgstr "ne peut être vide"
+  ```
   """
-  def error_tag(form, field) do
-    Enum.map(Keyword.get_values(form.errors, field), fn error ->
-      content_tag(:span, translate_error(error))
+  def error_messages(changeset) do
+    changeset
+    |> Changeset.traverse_errors(&translate_error/1)
+    |> convert_errors_to_html(changeset.data.__struct__)
+  end
+
+  defp translate_error({message, options}) do
+    if options[:count] do
+      Gettext.dngettext(ElixirBoilerplate.Gettext, "ecto", message, message, options[:count], options)
+    else
+      Gettext.dgettext(ElixirBoilerplate.Gettext, "ecto", message, options)
+    end
+  end
+
+  defp convert_errors_to_html(errors, schema) do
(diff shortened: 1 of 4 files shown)


Boubalou:
> I added basic tests but I’ll keep on adding more before merging because it doesn’t fully support “many” fields like has_many and embeds_many associations.

Let us know once this is feature complete, I'll dive into it then.

remiprev:
@Boubalou Check it out! I added more tests and it actually helped fix an existing bug 😬

lib/elixir_boilerplate_web/errors/helpers.ex
```diff
@@ -1,43 +1,48 @@
 defmodule ElixirBoilerplateWeb.Errors.Helpers do
-  @moduledoc """
```
Boubalou: Any particular reason to remove the `@moduledoc`?

remiprev: Yes, since the module now exposes a single function (`error_messages/1`) and that function is documented (with `@doc`), I couldn’t come up with a relevant `@moduledoc` :smile:

lib/elixir_boilerplate_web/errors/helpers.ex
```diff
@@ -1,43 +1,47 @@
...
-    #
-    #     # Translate the number of files with plural rules
-    #     dngettext("errors", "1 file", "%{count} files", count)
-    #
-    # Because the error messages we show in our forms and APIs
-    # are defined inside Ecto, we need to translate them dynamically.
-    # This requires us to call the Gettext module passing our gettext
-    # backend as first argument.
-    #
-    # Note we use the "errors" domain, which means translations
```
Boubalou: I was about to ask why we renamed the "errors" domain to "ecto" but figured it made more sense. But since these comments were removed, i feel that we may lack the visibility of the expectations here... which is to have an "ecto.po" file in the translations.

lib/elixir_boilerplate_web/errors/helpers.ex
```diff
@@ -1,43 +1,47 @@
...
+  end
+
+  defp convert_error_field({field, errors}, memo, schema) when is_list(errors), do: memo ++ Enum.flat_map(errors, &convert_error_subfield(&1, field, [], schema))
+  defp convert_error_field({field, errors}, memo, schema) when is_map(errors), do: memo ++ Enum.flat_map(Map.keys(errors), &convert_error_subfield(&1, field, errors[&1], schema))
+
+  defp convert_error_subfield(message, field, _, _schema) when is_binary(message) do
+    # NOTE `schema` is available here if we want to use something like
+    # `schema.humanize_field(field)` to be able to display `"Email address is
+    # invalid"` instead of `email is invalid"`.
+    ["#{field} #{message}"]
```
Boubalou: I like the idea here. It would be nice to cover this as an example through tests.

lib/elixir_boilerplate_web/errors/helpers.ex
remiprev: You’re right, I’ll add the explanation in `error_messages/1`’s `@doc` block.

lib/elixir_boilerplate_web/errors/helpers.ex
remiprev: You mean the example using a `schema.humanize_field/1` function?

It would be kind of hard to _test_ it because we would have to actually use it here in the helper 😕

PR #57, opened 2019-06-13
Add WithClauses Credo check rules

romarickb:
* Update Credo to version `1.1`
* Add `WithClauses` checks in `.credo.exs`


romarickb:
Sometimes we have PR comments on use of `with` clause.
This check can avoid to repeat this kind of comments : 
* Do not start or end with non-pattern-matching clauses;
* Do not use only one pattern matching clause with `else`.

Take a look at two examples from the [Credo.Check.Refactor.WithClauses module](https://github.com/rrrene/credo/blob/master/lib/credo/check/refactor/with_clauses.ex).

PR #59, opened 2019-06-18
Add “mix check.unused_locked_deps” task

remiprev:
We want to avoid keeping unused dependencies in `mix.lock`. This pull request adds a new `check.unused_locked_deps` that is used in the CI build.

I also renamed our other check `erlang.check_version` to `check.erlang_version` in order to keep things consistent 😄

Also, there were three unused deps in `mix.lock` 🙄


lib/mix/tasks/check.erlang_version.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Mix.Tasks.Erlang.CheckVersion do
+defmodule Mix.Tasks.Check.ErlangVersion do
```
gcauchon: Consistency 🥇

PR #61, opened 2019-06-18
Remove Dialyzer from the project

remiprev:
I think we started using (and enforcing) Dialyzer a little too early in our boilerplate code.

So far, we’re spending time ignoring and fixing typespec errors from external packages instead of _saving_ time by catching bugs early in development thanks to typespecs.

We might need more time to evaluate it more thoroughly and what value it can add to our projects. _Then_, we’ll be able to settle on its place in our boilerplate project.

What do you think?


thermech:
👍

PR #62, opened 2019-06-20
Add “AvoidSpecificTermsInModuleNames” check

remiprev:
Now that we’re using [`credo_naming`](https://github.com/mirego/credo_naming), we can take advantage of its other checks.

This pull requests adds the `CredoNaming.Check.Warning.AvoidSpecificTermsInModuleNames` check, which can prevent us from using OOP-style terms in our module names.

Here is the list of terms I’m suggesting we use for the moment:

* `Manager`
* `Fetcher`
* `Builder`
* `Persister`
* `Serializer`
* `Helper`
* `Helpers`

I think this will make it easier for us to stick to our [naming guidelines](https://github.com/mirego/elixir-boilerplate/blob/master/docs/module-naming.fr.md).

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -55,6 +55,7 @@
         {Credo.Check.Warning.UnusedTupleOperation},
         {Credo.Check.Warning.OperationWithConstantResult},
         {CredoEnvvar.Check.Warning.EnvironmentVariablesAtCompileTime},
+        {CredoNaming.Check.Warning.AvoidSpecificTermsInModuleNames, terms: ["Manager", "Fetcher", "Builder", "Persister", "Serializer", "Helper", "Helpers", "Utils"]},
         {CredoNaming.Check.Consistency.ModuleFilename, excluded_paths: ["config", "mix.exs", "priv", "test/support"]}
       ]
     }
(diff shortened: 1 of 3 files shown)


.credo.exs
```diff
@@ -55,6 +55,7 @@
         {Credo.Check.Warning.UnusedTupleOperation},
         {Credo.Check.Warning.OperationWithConstantResult},
         {CredoEnvvar.Check.Warning.EnvironmentVariablesAtCompileTime},
+        {CredoNaming.Check.Warning.AvoidSpecificTermsInModuleNames, terms: ["Manager", "Fetcher", "Builder", "Persister", "Serializer", "Helper", "Helpers", "Utils"]},
```
Boubalou: Is there a way to use a regex to cover all use cases with ending `s`? It would be nice :)

remiprev: The check doesn’t support regular expressions yet but this is a great feature idea. I will implement it soon 👍

remiprev: @Boubalou Your feature is ready! 😉 

https://github.com/mirego/credo_naming/pull/4

PR #65, opened 2019-07-16
Add ExMachina blank setup

### LINKED ISSUE #5: Add ex_machina factory setup
* We’re using [ExMachina](https://github.com/thoughtbot/ex_machina) in the majority of our Elixir projects that use Ecto
* Ecto is included in this boilerplate project

Therefore, a basic (working) ExMachina setup should be included in this boilerplate project 😄

remiprev:
This pull request adds a very basic ExMachina setup.

Fixes #5


remiprev:
> Should we add a factory module and show how to use it in an existing test (healthcheck plug maybe?) to complete the setup.

Yeah, I thought about adding a factory but I don’t want to have to bundle an Ecto schema into the project. I could add a factory that would build a simple struct but that won’t take advantage of the `ExMachina.Ecto` integration.

> Furthermore, `Faker` is useful to generate fake data in our factories, we could add it to the `:dev, :test` dependencies.

Yes, good point. I’ll add it too!

test/support/factory.ex
```diff
@@ -0,0 +1,8 @@
+defmodule ElixirBoilerplate.Factory do
+  use ExMachina.Ecto, repo: ElixirBoilerplate.Repo
+
+  # This is a sample factory to make sure our setup is working correctly.
+  def name_factory(_) do
```
simonprev: No need for the arity 1. By default we should use arity 0. The arity 1 function is the "full control" mode https://github.com/thoughtbot/ex_machina/#full-control-of-factory

remiprev: No we have to use the arity 1 function because we don’t want to return a map. Factories with arity 0 functions must return a map.

remiprev: If we use an arity 0 function:

PR #69, opened 2019-07-30
Make the majority of environment variables required for dev and prod

remiprev:
Since we provide all required environment variables in `.env.dev` and most of them are required for the application to start, we could actually _enforce_ them in order to get an explicit error message if one of them is missing.

@simonprev @gcauchon @paquetgp What do you think? Is it a terrible idea? Is there something missing?

### DIFF
--- a/.env.dev
+++ b/.env.dev
@@ -26,6 +26,7 @@ SIGNING_SALT= # Generate salt with `mix phx.gen.secret`
 # - Use `postgres://postgres:development@localhost/elixir_boilerplate_dev` if you’re using the PostgreSQL server provided by Docker Compose
 DATABASE_URL=postgres://localhost/elixir_boilerplate_dev
 DATABASE_POOL_SIZE=20
+DATABASE_SSL=false
 
 # Basic Authentication
 # BASIC_AUTH_USERNAME=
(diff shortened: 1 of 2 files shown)


simonprev:
Does it affect only `releases`? Can I still do `mix deps.get` without those variables?

gcauchon:
I’m not sure the optional keyword list is the best "API" to express mandatory environment variables. 

_Bang_ `!` fonctions that _raise_ if the key is missing might be more explicit?

remiprev:
> I’m not sure the optional keyword list is the best "API" to express mandatory environment variables.
> 
> _Bang_ `!` fonctions that _raise_ if the key is missing might be more explicit?

Yep, that makes sense! 👍

gcauchon:
The "new" `System` functions in `1.9` are achieving the same goal:

-  [`System.fetch_env/1`](https://hexdocs.pm/elixir/System.html#fetch_env/1)
-  [`System.fetch_env!/1`](https://hexdocs.pm/elixir/System.html#fetch_env!/1)

I'm quoting new because @simonprev made me realize there are simply following the naming convention of other collections/structs from the standard lib.

remiprev:
> Does it affect only `releases`? Can I still do `mix deps.get` without those variables?

Yep, no… I’m gonna close this.

PR #64, opened 2019-07-12
Add Sobelow and wire it to the CI pipeline

gcauchon:
## 📚 Context

I first heard of `Sobelow` while listening [episode 056](https://devchat.tv/elixir-mix/emx-056-sobelow-and-security-with-griffin-byatt/) of the Elixir Mix podcast : _Sobelow and Security with_ @GriffinMB (author of the library). I was curious to see the report for a few of our Elixir apps running in production; I instantly was able to fix a few quirks to better "protect" our systems and our users.

This brings back the basic config and CI pipeline check to the boilerplate project!

## 📷 Result

**Before**

```shell
> mix sobelow --exit --ignore Config.HTTPS
##############################################
#                                            #
#          Running Sobelow - v0.8.0          #
#  Created by Griffin Byatt - @griffinbyatt  #
#     NCC Group - https://nccgroup.trust     #
#                                            #
##############################################

Config.CSP: Missing Content-Security-Policy - High Confidence
File: lib/elixir_boilerplate_web/router.ex
Pipeline: browser
Line: 13

-----------------------------------------------

... SCAN COMPLETE ...
```

**After**

```shell
> make check-code-security
mix sobelow --config
Checking Sobelow version...

##############################################
#                                            #
#          Running Sobelow - v0.8.0          #
#  Created by Griffin Byatt - @griffinbyatt  #
#     NCC Group - https://nccgroup.trust     #
#                                            #
##############################################

... SCAN COMPLETE ...
```

# 🎉

PR #70, opened 2019-07-30
Improve Content-Security-Policy value

remiprev:
We implemented a default `Content-Security-Policy` value in #64 but it did added a lot for console errors in development.

It also prevented the demo homepage from loading the project logo from `user-images.githubusercontent.com`.

So now we implement some default directives for script and styles that are compatible with Phoenix development stuff and we replaced the logo image for a simple text logo.

I also had to disable the `Config.CSP` Sobelow check because we put the header in a separate plug.

### DIFF
--- a/.sobelow-conf
+++ b/.sobelow-conf
@@ -7,6 +7,6 @@
   format: "txt",
   out: "",
   threshold: "low",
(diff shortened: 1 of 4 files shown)


remiprev:
Closed in favor of #74

PR #71, opened 2019-08-19
Ditch deprecated Supervisor.Spec module

remiprev:
We were still using helper functions from [`Supervisor.Spec`](https://hexdocs.pm/elixir/Supervisor.Spec.html) to build the main supervisor tree.

> This module is deprecated. Use the new child specifications outlined in the Supervisor module instead.
> 
> Outdated functions for building child specifications.
> 
> The functions in this module are deprecated and they do not work with the module-based child specs introduced in Elixir v1.5. Please see the Supervisor documentation instead.
> 
> Convenience functions for defining supervisor specifications.

### DIFF
--- a/lib/elixir_boilerplate/application.ex
+++ b/lib/elixir_boilerplate/application.ex
@@ -4,15 +4,11 @@ defmodule ElixirBoilerplate.Application do
   """
 
   use Application
-  alias ElixirBoilerplate.Repo
-  alias ElixirBoilerplateWeb.Endpoint
 
   def start(_type, _args) do
-    import Supervisor.Spec
-
     children = [
-      supervisor(Repo, []),
-      supervisor(Endpoint, [])
+      ElixirBoilerplate.Repo,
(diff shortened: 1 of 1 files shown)


lib/elixir_boilerplate/application.ex
```diff
@@ -8,12 +8,7 @@ defmodule ElixirBoilerplate.Application do
   alias ElixirBoilerplateWeb.Endpoint
 
   def start(_type, _args) do
-    import Supervisor.Spec
-
-    children = [
-      supervisor(Repo, []),
-      supervisor(Endpoint, [])
-    ]
+    children = [Repo, Endpoint]
```
gcauchon: We could even remove the aliases:

```suggestion
    children = [
      ElixirBoilerplate.Repo,
      ElixirBoilerplateWeb.Endpoint
    ]
```

remiprev: Yep, much cleaner! 👌

PR #73, opened 2019-08-21
Ensure `:ssl` is started before Ecto.Migrator is ran

remiprev:
If an `Ecto.Repo` is configured with `ssl: true`, we have to make sure the `:ssl` application is started before `Ecto.Migrator` is used (eg. when the application is started).


remiprev:
@simonprev I think it’s safe to start `:ssl` before migrations are ran _even_ if `DATABASE_SSL` is not enabled. What do you think?

simonprev:
Sure, the `ssl` app is included anyway in the app, there is no "risk" in starting it for non-ssl database. 👍

PR #77, opened 2019-09-10
Update dependencies + 3 steps Dockerfile

gcauchon:
1. To keep the boilerplate up-to-date, let’s update dependencies for both Elixir and Node.

2. I also made a gap analysis with a vanilla project created with `mix phx.new`; a few things were out of line so I updated those.

3. To simplify the Docker setup, I also added an initial build step to bundle static assets using Webpack in a dedicated image instead of adding `node` to the `elixir:1.9-alpine` one.


Dockerfile
```diff
@@ -1,5 +1,26 @@
...
+FROM node:10.16-alpine AS js-builder
+
+ARG NODE_ENV=prod
+
+ENV NODE_ENV=${NODE_ENV}
+
+WORKDIR /build
+
+# Install Alpine dependencies
+RUN apk update --no-cacje && \
```
garno:
```suggestion
RUN apk update --no-cache && \
```

Dockerfile
```diff
@@ -13,42 +34,43 @@ ENV APP_NAME=${APP_NAME} \
 
 WORKDIR /build
 
-# This step installs all the build tools we'll need
-RUN apk update && \
+# Install Alpine dependencies
+RUN apk update --no-cacje && \
```
garno:
```suggestion
RUN apk update --no-cache && \
```

PR #78, opened 2019-09-20
Fix empty Absinthe context

garno:
For reason unknown to my knowledge, using the context plug after `dispatch` prevent the context from being populated. The order has been [changed recently](https://github.com/mirego/elixir-boilerplate/commit/6747a2a4e38376d88277b12ff64b74fdfc5064d3#diff-8718fe63175e93ad62609310fa049129R12).

I’ve read the dispatch plug code and couldn’t understand why 🤷‍♂


garno:
> I’ll try to understand what’s wrong when I have a few minutes…

Keep me posted, I’m really intrigued.

PR #81, opened 2019-10-23
Add a new makefile's command to easily start the server

Boubalou:
## 📖 Description

To ease the process of starting the Phoenix server, let's use the makefile to isolate the knowledge by exposing a `run` command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,10 @@ push: ## Push the Docker image
 # Development targets
 # -------------------
 
+.PHONY: run
+run: ## Run the server as IEX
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -60,6 +60,10 @@ push: ## Push the Docker image
 # Development targets
 # -------------------
 
+.PHONY: run
+run: ## Run the server as IEX
```
remiprev: 😉

```suggestion
run: ## Run the server inside an IEx shell
```

PR #85, opened 2020-01-08
Add data-testid attributes in HTML template files

romarickb:
In order to test return data in HTML controllers we need to add "data-testid" attributes to the html tags. 
```elixir
<p <%= testid_attribute("description-paragraph") %>>This repository is the stable base upon which we build our Elixir projects at Mirego.<br />We want to share it with the world so you can build awesome Elixir applications too.</p>
<%= content_tag :p, data_testid: testid_key("message-paragraph") do %>
    <strong>Message:</strong> <%= @message %>
<% end %>
```

However, we do not want these attributes to be found in the release. So we can configure in which environnement this tags will be available.
- `config :elixir_boilerplate, ElixirBoilerplateWeb, data_identity_provider: ElixirBoilerplateWeb.DataIdentifier.TestID` if we want `data-testid` attributes to be rendered
- `config :elixir_boilerplate, ElixirBoilerplateWeb, data_identity_provider: ElixirBoilerplateWeb.DataIdentifier.NoTestID` if we don't want `data-testid` attributes to be rendered


config/dev.exs
```diff
@@ -21,6 +21,8 @@ config :elixir_boilerplate, ElixirBoilerplateWeb.Endpoint,
     ]
   ]
 
+config :elixir_boilerplate, ElixirBoilerplateWeb, data_identity_provider: ElixirBoilerplateWeb.DataIdentifier.TestID
```
gcauchon: Could we move this to `config/config.exs` and only overwrite it in `config/prod.exs`?

lib/elixir_boilerplate_web/home/templates/index.html.eex
```diff
@@ -3,6 +3,8 @@
...
   </a>
 
-  <p>This repository is the stable base upon which we build our Elixir projects at Mirego.<br />We want to share it with the world so you can build awesome Elixir applications too.</p>
-  <p><strong>Message:</strong> <%= @message %></p>
+  <p <%= testid_attribute("description-paragraph") %>>This repository is the stable base upon which we build our Elixir projects at Mirego.<br />We want to share it with the world so you can build awesome Elixir applications too.</p>
+  <%= content_tag :p, data_testid: testid_key("message-paragraph") do %>
```
gcauchon: Doesn't this defy the purpose since `data-testid=""` will end up in the DOM in production?

romarickb: if an attribute has `nil` value it will not be added in resulting `p` html tag

romarickb:
I m closing this issue because we gonna use a library instead embedding it in boilerplate

PR #88, opened 2020-02-05
Scope docker COPY + hex 0.20.2+ compatible mix.lock

gcauchon:
[`hex 0.20.2`](https://github.com/hexpm/hex/blob/master/CHANGELOG.md#v0202-2020-02-03) was released yesterday which includes a fix for symlinks in tarballs and a new checksum for the whole archive.

https://elixirforum.com/t/docker-build-started-failling-today/28844

However, the fix had a breaking side effect in a Docker setup, like this one, where dependencies are installed first, then the codebase is copied with `COPY . .`, causing the Hex manifest to be overlayed.

@ericmj understands and explains it better than I could :

> It happens when you do `COPY . .` after `RUN mix deps.get`, because it copies everything from the host to the container, including `mix.lock` which is potentially outdated compared to the manifest in the container.

_Imho_, there are two important takeaways here:

1. Even with a well crafted `.dockerignore` file, `COPY . .` is often too inclusive to get adequate control on what needs to be built.
2. Even if [`hex 0.20.5`](https://github.com/hexpm/hex/blob/master/CHANGELOG.md#v0205-2020-02-05) is already available and supports backward compatibily, it is preferable to update `mix.lock` to include the _outer checksum_ asap!


Dockerfile
```diff
@@ -67,24 +74,23 @@ FROM alpine:3.10
...
-# Update Alpine and install dependencies
+# Install Alpine dependencies
 RUN apk update --no-cache && \
     apk upgrade --no-cache && \
     apk add --no-cache bash openssl-dev erlang-crypto
 
 WORKDIR /opt/elixir_boilerplate
 
-# Copy the OTP binary from the build step
+# Copy OTP release drom Step 2
```
romarickb: From step 2

Dockerfile
```diff
@@ -43,18 +45,23 @@ RUN apk update --no-cache && \
 RUN mix local.rebar --force && \
     mix local.hex --force
 
-# Copy dependency config first
+# Install & complie Hex dependencies
```
romarickb: Don’t you think dependencies is enough ? Because sometimes we could have dependencies not from Hex.

gcauchon: I wanted to follow the other comments format…

> ` Install … dependencies

Dockerfile
```diff
@@ -43,18 +45,23 @@ RUN apk update --no-cache && \
...
-# Copy codebase and compile
-COPY . .
+RUN mix deps.get --only ${MIX_ENV} && \
+    mix deps.compile
+
+# Compile codebase
+COPY config config
+COPY lib lib
+COPY priv priv
+COPY rel rel
```
remi: Do we need the `rel` directory to run `mix compile` or could we only copy it just before running `mix release`?

gcauchon: Excellent question; I’ll give it a try!

My guess is your right, we could move it down before `mix release`…

gcauchon: I does work 🥇

PR #91, opened 2020-02-11
Moving from Travis to GitHub Actions

gcauchon:
Title says it all; we are moving from Travis to leverage Github Actions 💃🕺


gcauchon:
@romarickb I added you to the reviewers list because I know you did a similar transition for a customer project. Let me know if you did things differently; especially regarding the `npm` cache and installing JS dependencies.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,48 @@
...
+        id: cache-npm
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-npm-${{ hashFiles(format('{0}/assets/package-lock.json', github.workspace)) }}
+          restore-keys: |
+            ${{ runner.os }}-npm-
+      - run: make dependencies 
+      - run: make check
+      - run: make lint
+      - run: make test
```
Boubalou: Shall we not keep using the script `ci-check.sh` instead of each step in different runs like that?

Madumo: Individual steps makes a way better output in the github action UI. Plus, you know in a glimpse what failed or not. Our ci-check script was great with travis which doesn't have a great UI for outputs, but isn't necessary with github actions.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,48 @@
...
+          key: ${{ runner.os }}-mix-${{ hashFiles(format('{0}/mix.lock', github.workspace)) }}
+          restore-keys: |
+            ${{ runner.os }}-mix-
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12.15
+      - uses: actions/cache@v1
+        id: cache-npm
+        with:
+          path: ~/.npm
```
Madumo: I had better success speed wise caching the node_modules directory in my tests, but i don't know how that might turn bad. It was also using yarn, maybe npm handles its cache folder in a better way. But you should try and compare! :)

.github/workflows/ci.yaml
gcauchon: _^this_ is my take on the _future_ of `ci-checks.sh` too!

In the end, it would be kept for _development workflow_ purposes, but I personally think the developer "APIs" should all be in `make` and we already maintain the `check` target for that matter…

Let's see what others think before removing it…

.github/workflows/ci.yaml
gcauchon: Following `actions/cache@v1` documentation:

> For npm, cache files are stored in ~/.npm on Posix, or %AppData%/npm-cache on Windows. See https://docs.npmjs.com/cli/cache#cache
>
> > _Note_: It is not recommended to cache `node_modules`, as it can break across Node versions and won't work with `npm ci`

https://github.com/actions/cache/blob/master/examples.md#node---npm

.github/workflows/ci.yaml
gcauchon: Summoning @simonprev @remi @garno @ging @lucbernard…

Makefile
```diff
@@ -75,7 +75,7 @@ run: ## Run the server inside an IEx shell
 .PHONY: dependencies
 dependencies: ## Install dependencies
 	mix deps.get --force
-	npm install --prefix assets
+	npm ci --prefix assets
```
simonprev: This is great for CI but not for development, the developer should use `npm install`, amd the ci `npm ci`

gcauchon: True, I'll see what I could come up with 🤔

Makefile
```diff
@@ -56,6 +56,11 @@ targets:
 # Build targets
 # -------------
 
+.PHONY: prepare
+prepare: ## Install dependencies
```
gcauchon: @simonprev I added a `prepare` target in the "Build" scope, but I'm not sure of the naming convention…

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,49 @@
+name: Run codebase checks and tests
+
+on: [push, pull_request]
+
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+
+    services:
+      db:
```
romarickb: I think you should had 
```
db:
 env:
    POSTGRES_USER: postgres
    POSTGRES_PASSWORD: postgres
    POSTGRES_DB: postgres
```

Or you gonna have the following error : `Failed to initialize, db service is unhealthy.`

Makefile
```diff
@@ -56,6 +56,11 @@ targets:
...
+	mix deps.get
+	npm ci --prefix assets
```
romarickb: Should we not separate elixir dependencies and node dependencies to be able to cache and execute step independently regarding if cache exists ?

We could have in `ci.yml` something like 
```
- if: steps.cache-mix.outputs.cache-hit != 'true'
   run: make dependencies
```

gcauchon: I don’t think we need to check if there was a cache hit or not. In the end, after restoring the caches, `mix deps.get` will simply run is a few miliseconds!

As for `npm`, as mentionned above, the recommanded caching uses a different approach following `actions/cache@v1` documentation:

> For npm, cache files are stored in ~/.npm on Posix, or %AppData%/npm-cache on Windows. See https://docs.npmjs.com/cli/cache#cache
>
> _Note_: It is not recommended to cache `node_modules`, as it can break across Node versions and won't work with `npm ci`

https://github.com/actions/cache/blob/master/examples.md#node---npm

.github/workflows/ci.yaml
gcauchon: It works with specific environement variables, but it might be related to the healthcheck options.

https://github.com/mirego/elixir-boilerplate/pull/91/files#diff-60fe268f71ae7f8409aff40d0a3e51d8R20

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,65 @@
...
+          path: ~/.npm
+          key: ${{ runner.os }}-npm-${{ hashFiles(format('{0}/assets/package-lock.json', github.workspace)) }}
+          restore-keys: |
+            ${{ runner.os }}-npm-
+
+      - run: make prepare
+      - run: make check
+      - run: make lint
+      - run: make test
+
```
remi:
```suggestion
```

PR #92, opened 2020-02-11
Add an experimental local database URL check for the test configuration

alarochelle:
In order to prevent mistakes when running tests in an application, we add an experimental local database URL check in the `test.exs` configuration file.

If the specified `DATABASE_URL` used to run tests is not a local URL, an error will be raised, thus preventing the tests from running and from potentially damaging the database.

This change is experimental in the sense that it may not be _necessary_, but could be very useful in some cases.


config/releases.exs
```diff
@@ -31,6 +31,24 @@ defmodule Environment do
...
+      %{host: "localhost"} ->
+        url
+
+      %{host: "127.0.0.1"} ->
+        url
+
+      %{host: nil} ->
+        url
+
+      %{host: _} ->
```
remi: Since we’re using this pattern as a catch-all, we could to:

```suggestion
      _ ->
```

config/releases.exs
```diff
@@ -31,6 +31,24 @@ defmodule Environment do
...
+        raise "Make sure that the DATABASE_URL environment variable for the tests points to a local database."
```
Boubalou: The method look re-usable for any URL but the error it raises is bound to DATABASE_URL.

config/releases.exs
```diff
@@ -31,6 +31,24 @@ defmodule Environment do
       _ -> nil
     end
   end
+
+  def get_local_url(key) do
```
simonprev: I think the function could be a bit smaller and the error message more helpful:

```elixir
url = get(key)

case URI.parse(url).host do
  host when host in ~w(localhost 127.0.0.1) -> url
  host -> raise "Expected host of the DATABASE_URL environment variable to be localhost or 127.0.0.1, got: #{host}"
end
```

config/releases.exs
```diff
@@ -31,6 +31,24 @@ defmodule Environment do
...
+    url = get(key)
+
+    case URI.parse(get(key)) do
```
remi: Since we already got it:

```suggestion
    case URI.parse(url) do
```

config/releases.exs
Boubalou: Maybe explicitly rename it to `get_database_url()`, what do you think?

remi: I think we just need to rephrase the message and make it a bit more generic:

```suggestion
        raise "Make sure that the #{key} environment variable points to a local URL."
```

config/releases.exs
remi: If we’re going with this, I suggest

```elixir
  host -> raise "Expected host of the #{key} environment variable to be `localhost` or `127.0.0.1`, got: #{host}"
```

Boubalou:
I agree with the initiative 💯 !

config/releases.exs
alarochelle: I think we should go with what @simonprev and @remi proposed which is to make the error message more generic by using the `key` and `host` parameters.

Boubalou: Yes!

config/releases.exs
```diff
@@ -31,6 +31,15 @@ defmodule Environment do
       _ -> nil
     end
   end
+
+  def get_local_url(key) do
+    url = get(key)
+
+    case URI.parse(url).host do
+      host when host in ~w(localhost 127.0.0.1) -> url
```
charlesdemers: Should we allow `0.0.0.0` and `host.docker.internal`?

alarochelle: You're right, they should probably be allowed as there is no risk of doing so anyways. What do others think about this?

config/releases.exs
```diff
@@ -31,6 +31,15 @@ defmodule Environment do
       _ -> nil
     end
   end
+
+  def get_local_url(key) do
+    url = get(key)
+
+    case URI.parse(url).host do
+      host when host in ~w(0.0.0.0 127.0.0.1 localhost host.docker.internal) -> url
+      host -> raise "Expected host of the #{key} environment variable to be `localhost` or `127.0.0.1`, got: #{host}"
```
romarickb: Maybe we could use a varaible for  `~w(0.0.0.0 127.0.0.1 localhost host.docker.internal)` and use it to be more precise in the raise message exception

remi: True, it’s a little misleading to provide two valid values when we actually have four valid ones 😄

PR #94, opened 2020-03-26
Improve “npx” usage in Makefile

remi:
In our other boilerplate projects ([Ember.js](https://github.com/mirego/ember-boilerplate) and [React](https://github.com/mirego/react-boilerplate)), we can use `npx` to run our linting and formatting tools because the `package.json` file is at the project root.

However, in this project, it is in `assets`. And `npx` currently doesn’t support passing a prefix to it (unline `npm --prefix assets …`). It only supports going _up_ in the directory tree to look for the project root. So running `npx prettier` _installs_ `prettier`, it doesn’t use our locked version in `assets/package.json`.

I tried messing around with `$PATH` to no avail:

```
$ PATH="./assets/node_modules/.bin:$PATH" npx --no-install prettier --version
not found: prettier

$ PATH="./assets:$PATH" npx --no-install prettier --version
not found: prettier
```

So I think for now, we should avoid using `npx` when executing commands from the project root.

I also took the time to upgrade ESLint to 6.x, because since we were already using `npx eslint`, our setup was ready for it (and actually didn’t work ESLint 5.x 🤦‍♂)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,8 @@ DOCKER_REMOTE_IMAGE = $(DOCKER_REGISTRY)/$(DOCKER_LOCAL_IMAGE)
 # Linter and formatter configuration
 # ----------------------------------
 
-PRETTIER_FILES_PATTERN = 'assets/.babelrc' 'assets/webpack.config.js' 'assets/{js,css,scripts}/**/*.{js,graphql,scss,css}' '**/*.md'
-STYLES_PATTERN = 'assets/css'
+PRETTIER_FILES_PATTERN = '.babelrc' 'webpack.config.js' '{js,css,scripts}/**/*.{js,graphql,scss,css}' '**/*.md'
+STYLES_PATTERN = 'css'
 
 # Introspection targets
 # ---------------------
@@ -107,7 +107,7 @@ check-code-security:
 .PHONY: check-format
 check-format:
 	mix format --dry-run --check-formatted
-	npx prettier --ignore-path assets/.prettierignore --check $(PRETTIER_FILES_PATTERN)
+	cd assets && npx prettier --check $(PRETTIER_FILES_PATTERN)
 
 .PHONY: check-unused-dependencies
 check-unused-dependencies:
@@ -116,7 +116,7 @@ check-unused-dependencies:
 .PHONY: format
 format: ## Format project files
 	mix format
-	npx prettier --ignore-path assets/.prettierignore --write $(PRETTIER_FILES_PATTERN)
+	cd assets && npx prettier --write $(PRETTIER_FILES_PATTERN)
 
 .PHONY: lint
 lint: lint-elixir lint-scripts lint-styles ## Lint project files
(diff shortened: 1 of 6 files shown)


assets/package.json
```diff
@@ -8,11 +8,9 @@
     "npm": "^6.13"
   },
   "scripts": {
-    "deploy": "npm run enforce-engine-versions && npx webpack --mode production",
-    "watch": "npm run enforce-engine-versions && npx webpack --mode development --watch-stdin",
-    "enforce-engine-versions": "./scripts/enforce-engine-versions.js",
-    "preinstall": "npm run enforce-engine-versions",
-    "postinstall": "npm run enforce-engine-versions"
```
gcauchon: These were redundant?

remi: Oops, that was an experiment 😄

Makefile
```diff
@@ -107,7 +107,7 @@ check-code-security:
 .PHONY: check-format
 check-format:
 	mix format --dry-run --check-formatted
-	npx prettier --ignore-path assets/.prettierignore --check $(PRETTIER_FILES_PATTERN)
+	assets/node_modules/.bin/prettier --ignore-path assets/.prettierignore --check $(PRETTIER_FILES_PATTERN)
```
charlesdemers: Why not using `cd` to change the directory before executing the script?

```suggestion
	cd assets && npx prettier --ignore-path .prettierignore --check $(PRETTIER_FILES_PATTERN)

Comment (charlesdemers) [reply Makefile]: And in fact, we could also remove the `--ignore-path` flag since we would already be in the right folder! 🤯

Comment (pboutin) [reply Makefile]: to avoid `cd` messing with the user's current folder: 
`(cd assets; npx prettier --ignore-path .prettierignore --check $(PRETTIER_FILES_PATTERN))`

Comment (charlesdemers) [reply Makefile]: Makefiles aren’t shell scripts, they execute in a separate shell and therefore don’t mess with the current directory in which you execute the script

Comment (remi) [reply Makefile]: You’re right, it wouldn’t mess with the user path, but I thought best to avoid it to prevent this:

```make
format:
	mix format                          # Working directory is `.`
	cd assets && npx prettier --write … # Working directory is `./assets`
    other-formatting-tool               # Working directory is still `./assets`
```

I guess we could use:

```make
cd assets && npx prettier --write … && cd ..
```

Even if `npx prettier --write …` fails and the second `&&` is skipped, it wouldn’t matter because the `other-formatting-tool` wouldn’t get executed anyway.

What do you guys think?

charlesdemers: In fact each line is executed separately, I tested it and given a target like this:

```
.PHONY: lol
lol:
	cd app && pwd
	pwd
```

I have the output:

```
❯ /Volumes/Mirego/projects/my-project
❯ make lol
cd app && pwd
/Volumes/Mirego/projects/my-project/app
pwd
/Volumes/Mirego/projects/my-project
```

remi: Oh, awesome then. Reverting to `cd assets && npx …`! 🎉

Makefile
```diff
@@ -128,8 +128,8 @@ lint-elixir:
 
 .PHONY: lint-scripts
 lint-scripts:
-	npx eslint --config assets/.eslintrc --ignore-path assets/.eslintignore --resolve-plugins-relative-to assets assets
+	cd assets && npx eslint
 
 .PHONY: lint-styles
 lint-styles:
-	npx stylelint --syntax scss --config assets/.stylelintrc $(STYLES_PATTERN)
+	cd assets && nox stylelint --syntax scss $(STYLES_PATTERN)
```
charlesdemers:
```suggestion
	cd assets && npx stylelint --syntax scss $(STYLES_PATTERN)

PR #98, opened 2020-03-26
Add a deeper check for DB URL to enforce running against a test DB

Boubalou:
## 📖 Description

Following the change from @alarochelle in #92, I propose to push the check a little deeper.

While the initial proposal in #92 protects us against erasing production data, the latter protect our local data.

This way, if, by mistake, you do the classic `nv .env.dev.local make test` instead of `nv .env.test.local make test`, it will raise the following:

```bash
** (RuntimeError) Expected database URL to ends with '_test', got: postgres://localhost/elixir_boilerplate_dev
    config/test.exs:14: TestEnvironment.enforce_test_database_name/1
    (stdlib 3.12) erl_eval.erl:680: :erl_eval.do_apply/6
    (stdlib 3.12) erl_eval.erl:888: :erl_eval.expr_list/6
    (stdlib 3.12) erl_eval.erl:240: :erl_eval.expr/5
    (stdlib 3.12) erl_eval.erl:232: :erl_eval.expr/5
    (stdlib 3.12) erl_eval.erl:233: :erl_eval.expr/5
    (stdlib 3.12) erl_eval.erl:888: :erl_eval.expr_list/6
    (stdlib 3.12) erl_eval.erl:411: :erl_eval.expr/5
make: *** [test] Error 1
```

## 🦀 Dispatch
		
- `#dispatch/elixir`

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -3,6 +3,19 @@ import Config
 # Import runtime configuration
 import_config "releases.exs"
 
+defmodule TestEnvironment do
+  @database_name_suffix "_test"
+
+  def enforce_test_database_name(key) do
+    url = Environment.get_local_url(key)
+
+    case String.contains?(url, @database_name_suffix) do
+      true -> url
+      _ -> raise "Expected database URL to ends with '#{@database_name_suffix}', got: #{url}"
+    end
+  end
+end
+
 config :elixir_boilerplate, ElixirBoilerplateWeb.Endpoint,
   http: [port: 4001],
   server: false,
@@ -24,4 +37,4 @@ config :logger, level: :warn
 
 config :elixir_boilerplate, ElixirBoilerplate.Repo,
   pool: Ecto.Adapters.SQL.Sandbox,
-  url: Environment.get_local_url("DATABASE_URL")
+  url: TestEnvironment.enforce_test_database_name("DATABASE_URL")
--- a/config/test.exs
+++ b/config/test.exs
@@ -9,9 +9,10 @@ defmodule TestEnvironment do
   def enforce_test_database_name(key) do
     url = Environment.get_local_url(key)
 
-    case String.contains?(url, @database_name_suffix) do
-      true -> url
-      _ -> raise "Expected database URL to ends with '#{@database_name_suffix}', got: #{url}"
+    if String.contains?(url, @database_name_suffix) do
+      url
+    else
+      raise "Expected database URL to ends with '#{@database_name_suffix}', got: #{url}"
     end
   end
 end
--- a/config/test.exs
+++ b/config/test.exs
@@ -6,8 +6,8 @@ import_config "releases.exs"
 defmodule TestEnvironment do
   @database_name_suffix "_test"
 
-  def enforce_test_database_name(key) do
-    url = Environment.get_local_url(key)
+  def get_test_database_url do
+    url = Environment.get("DATABASE_URL")
 
     if String.contains?(url, @database_name_suffix) do
       url
@@ -38,4 +38,4 @@ config :logger, level: :warn
 
 config :elixir_boilerplate, ElixirBoilerplate.Repo,
   pool: Ecto.Adapters.SQL.Sandbox,
-  url: TestEnvironment.enforce_test_database_name("DATABASE_URL")
+  url: TestEnvironment.get_test_database_url()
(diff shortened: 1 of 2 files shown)


config/test.exs
```diff
@@ -3,6 +3,19 @@ import Config
...
+defmodule TestEnvironment do
+  @database_name_suffix "_test"
+
+  def enforce_test_database_name(key) do
+    url = Environment.get_local_url(key)
+
+    case String.contains?(url, @database_name_suffix) do
+      true -> url
+      _ -> raise "Expected database URL to ends with '#{@database_name_suffix}', got: #{url}"
+    end
```
gcauchon: We should use `if` since it’s a simple _boolean_ check. 

```suggestion
    if String.contains?(url, @database_name_suffix) do
      url
    else
      raise "Expected database URL to ends with '#{@database_name_suffix}', got: #{url}"
    end
```

remi:
I’m not sure we need _both_ approaches.

I think if we enforce the `_test` suffix, the “local hostname” check is redundant since if you happen to use a `*_test` database in a non-local hostname, you pretty much know what you’re doing.

gcauchon:
> I’m not sure we need _both_ approaches.
> 
> I think if we enforce the `_test` suffix, the “local hostname” check is redundant since if you happen to use a `*_test` database in a non-local hostname, you pretty much know what you’re doing.

I was actually asking myself the same question!

I not a big fan of over protection against _hypothetical_ user mistakes… A `production` database should never be locally writable! Is dropping your local database really an issue? It sucks, but we should have all the tools in place to put you back on your feet in a matter of seconds…

Boubalou:
The way I saw it when I proposed this protection is:

We are exposing a boilerplate for all to use, which means that not all setups will be the same and not all ways of building the infrastructure will be the same (i.e.: production DB not accessible locally). 

As for the cost of rebuilding your local DB, the same applies, often dummy seeds aren't maintained, your data might have a lot of value over time, etc...

Also, if we accept this change, I think both checks should be kept because #92 is built in a re-usable way for any environment keys, therefore has a lot of value for the future.

lucbernard:
Kind of feel like both checks should be kept. The way I see it, the checks should enforce the fact that tests are run using a DB that's both local AND disposable (*_test name).

To re-iterate over Mat's point, we all had it happen to us: wiping our local DB by mistake. Restoring our local DB _is_ possible, but makes us waste precious time. It's not uncommon to work on a project whose seed script is no longer up to date or simply does not work anymore. Even with working seeds, restoring them does not necessarily bring you back to the state you were working from before wiping your data.

remi:
> Kind of feel like both checks should be kept. The way I see it, the checks should enforce the fact that tests are run using a DB that's both local AND disposable (*_test name).

I still don’t get why we would need _both_ checks. If we enforce running tests against a `_test` database, why do we _also_ need to enforce running tests in a local database? The `_test` check is much more powerful.

Let’s see how it would work when running tests with various `DATABASE_URL` values:

```bash
# Would get caught with `_test` check
DATABASE_URL=postgres://user:pass@my-remote-server.com/project_production

# Would get caught with `_test` check
DATABASE_URL=postgres://localhost/project_development

# Would get caught with local host check
DATABASE_URL=postgres://user:pass@my-remote-server.com/project_test

# All good!
DATABASE_URL=postgres://localhost/project_test
```

The way I see it, we should not have something in place for the third use case — it doesn’t exist. If you’re running the test suite and providing a non-local URL with a database name that ends with `_test`, you _know_ what you’re doing.

Is there something I’m missing?

lucbernard:
@remi gotcha. I'd misunderstood the point of your earlier comments. I guess keeping only this check would indeed make sense since it would catch all _critical_ mistakes

gcauchon:
I think @remi proposition covers it all 👍

remi:
Well my proposition is to kinda revert #92 and use only _this_ check.

Boubalou:
> Well my proposition is to kinda revert #92 and use only this check.

Is not there some value to keep both as the way #92 was implemented, it is "re-usable" for other URL based environment key.

remi:
I don’t think so. #92 was made to address the same problem #98 fixes. #98 is just a lot simpler 😄

Boubalou:
Ready boys!

config/test.exs
```diff
@@ -3,6 +3,20 @@ import Config
 # Import runtime configuration
 import_config "releases.exs"
 
+defmodule TestEnvironment do
+  @database_name_suffix "_test"
+
+  def enforce_test_database_name(key) do
```
remi: I would be more explicit here:

```suggestion
  def get_test_database_url(key) do
```

PR #99, opened 2020-04-04
Improve Makefile targets

remi:
- [x] Add missing description for `make check` target
- [x] Remove unused `make clean` target
- [x] Remove description for internal `make prepare`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ targets:
 # -------------
 
 .PHONY: prepare
-prepare: ## Install dependencies
+prepare:
 	mix deps.get
 	npm ci --prefix assets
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -82,10 +82,6 @@ dependencies: ## Install dependencies
 	mix deps.get
 	npm install --prefix assets
 
-.PHONY: clean
-clean: ## Clean dependencies
```
gcauchon, line 86: I would personally keep `clean`, it is usefull when `check-unused-dependencies` raises…

remi: But it’s not documented nor used anywhere in the projet.

`make clean` is often used in `Makefile`s to remove temporary files or artifacts that were generated during the build or install, so I think the name is misleading.

I guess we could add `make dependencies-clean` to be more explicit but I don’t think it’s right to just add an abstract target to "hide" the `mix deps.clean --unused --unlock` command, especially when the command can be shown to the user by Mix in error/warning messages.

What do you think?

gcauchon: If `make check-unused-dependencies` explicitly prints the clean command I think this is perfect!

PR #104, opened 2020-04-10
Add documentation for /ping and /health routes

remi:
## 📖 Description

We have two greate routes that were undocumented: `GET /ping` and `GET /health`. I simply added the missing documentation 😄

@pmartel12 Since you’re the one who wrote the `PlugCheckup` implementation, do you think those tables documenting each individual check are valuable?


pmartel12:
> Since you’re the one who wrote the PlugCheckup implementation, do you think those tables documenting each individual check are valuable

Yes! 😄

PR #120, opened 2020-09-02
use Hex.Utils.otp_version/0 instead of ad hoc solution

adrianomitre:
## 📖 Description

Use [Hex.Utils.otp_version/0](https://github.com/hexpm/hex/blob/v0.20.5/lib/hex/utils.ex#L259) instead of re-reinventing the wheel.

## 📝 Notes

I opted for adding `hex_core` instead of the `dialyze`, from where the implementation was copied, because it is way lighter and do not suggest the use o dialyze for static analysis.

## 📓 References

In case one wonders why this is not implemented on [System](https://hexdocs.pm/elixir/master/System.html), I would suggest reading [System.otp_release now includes the full release number by eksperimental · Pull Request #8576 · elixir-lang/elixir](https://github.com/elixir-lang/elixir/pull/8576#issuecomment-450481677). It does not clarifies much, but give some historical perspective.


simonprev:
Hi Adrian, thanks for the pull request! I think we should do the same as `hex_core` and copy the function and reference the original implementation in a comment. No need to depend on `hex_core` for this simple function

gcauchon:
Personally, I don’t see how including a new lib (as core as it can be) to the dependency tree only to use a single function (which is actually copied from another library) "reinvents the wheel"!?

https://github.com/hexpm/hex/blob/v0.20.5/lib/hex/utils.ex#L258-L272

```elixir
# From https://github.com/fishcakez/dialyze/blob/6698ae582c77940ee10b4babe4adeff22f1b7779/lib/mix/tasks/dialyze.ex#L168
```

I also think we should copy the function from `dialize` and keep a reference to the source the same ways it was done in `hex_core`.

adrianomitre:
First of all, thanks for the quick reply.

"Reinventing the weel" may have been a bad choice of words. The motivation is avoiding replicating the solution via copy-and-paste, as it presents relevant maintainability challenges.

Do we all agree here that the best solution would be to move it to `System`? It could be via `System.otp_release/1` with a "resolution" parameter that defaults to `major`, so it falls back to current behavior? If you agree to support me, I could give it a shot and open a PR cousin to https://github.com/elixir-lang/elixir/pull/8576.

What are your thoughts?

adrianomitre:
For your appreciation https://github.com/elixir-lang/elixir/pull/10299

adrianomitre:
For your appreciation https://github.com/elixir-lang/elixir/pull/10299

PR #121, opened 2020-09-04
drop ineffectual flag from Makefile check-format

adrianomitre:
## 📖 Description

Drop ineffectual flag from Makefile `check-format`.

## 📝 Notes

The use of `--dry-run` combined with `--check-formatted` is ineffective and I would suggest using simply `mix format --check-formatted`.

## 📓 References

https://github.com/elixir-lang/elixir/blob/v1.10.4/lib/mix/lib/mix/tasks/format.ex#L428

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,7 +110,7 @@ check-code-security:
 
 .PHONY: check-format
 check-format:
-	mix format --dry-run --check-formatted
+	mix format --check-formatted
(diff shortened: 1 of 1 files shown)


remi:
You’re right, it’s not used. It’s great that you included the reference in your PR description 👌

remi:
Thank you Adriano! 🙏

PR #89, opened 2020-02-10
Elixir 1.10 + hexpm docker image

gcauchon:
Elixir `1.10` was released on January 27th with [a few nice enhancements](https://elixir-lang.org/blog/2020/01/27/elixir-v1-10-0-released/)!

In a related effort, the elixir/hex core team are [now building Docker images](https://elixirforum.com/t/yet-another-elixir-and-erlang-docker-image/28740) that are Elixir+Erlang+Alpine specific, and way more stable thant the "official" ones.

PR #126, opened 2020-10-08
Remove Check.ErlangVersion

gcauchon:
⚠️ This one might be controversial; see it as a proposition we could internally vote on!

## 📖 Description

Back in #89 when we upgrade to Elixir `1.10`, we also upgraded the base image for the ones maintained by the Hex team! I don’t see any value in maintaining an explicit check on the Erlang version anymore.

This would also close #120.


gcauchon:
As the check is at build time; the team decided to keep it as it validates the development setup and prevents unexpected errors caused by a misconfigured local environment.

PR #127, opened 2020-10-10
Fix warnings with Faker and Phoenix.ConnTest

Boubalou:
## 📖 Description

Fix deprecation warnings coming from Faker and the latest Phoenix upgrade as seen below:

```
warning: Using Phoenix.ConnTest is deprecated, instead of:

    use Phoenix.ConnTest

do:

    import Plug.Conn
    import Phoenix.ConnTest

  test/elixir_boilerplate_web/ping_test.exs:2: ElixirBoilerplateWeb.PingTest (module)
```

```
warning: Faker.Name.name/0 is deprecated. Use Faker.Person.name/0 instead.
  test/support/factory.ex:6: ElixirBoilerplate.Factory.name_factory/1
```

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -88,7 +88,7 @@ defmodule ElixirBoilerplate.Mixfile do
 
       # Test factories
       {:ex_machina, "~> 2.4", only: :test},
-      {:faker, "~> 0.12", only: :test},
+      {:faker, "~> 0.15", only: :test},
 
       # Test coverage
       {:excoveralls, "~> 0.13", only: :test}
--- a/test/support/conn_case.ex
+++ b/test/support/conn_case.ex
@@ -23,7 +23,9 @@ defmodule ElixirBoilerplateWeb.ConnCase do
   using do
     quote do
       # Import conveniences for testing with connections
-      use Phoenix.ConnTest
+      import Plug.Conn
+      import Phoenix.ConnTest
+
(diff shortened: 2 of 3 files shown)


mix.exs
```diff
@@ -88,7 +88,7 @@ defmodule ElixirBoilerplate.Mixfile do
 
       # Test factories
       {:ex_machina, "~> 2.4", only: :test},
-      {:faker, "~> 0.12", only: :test},
+      {:faker, "~> 0.15", only: :test},
```
Boubalou, line 91: This change doesn't impact the .lock because it was already generated with this version. 🤔

gcauchon (approved):

test/support/conn_case.ex
```diff
@@ -22,10 +22,12 @@ defmodule ElixirBoilerplateWeb.ConnCase do
 
   using do
     quote do
-      # Import conveniences for testing with connections
-      use Phoenix.ConnTest
       import ElixirBoilerplateWeb.Router.Helpers
```
remi, line 25: I think `ElixirBoilerplateWeb.Router.Helpers` should fall under the "conveniences for testing with connections" group.

Boubalou: I was unsure due to the alphabetic order thing since they all are imports now instead of `use` and `import`.

I will try it and see if our checks are good with it!

PR #125, opened 2020-10-08
Update dependencies

gcauchon:
## 📖 Description

Many updates were released in the last weeks/months, here a few `lib` updates; getting the boilerplate ready for Elixir `1.11` that will follow next!

- Erlang `23.1`
- Alpine `1.12`
- Ecto `3.5`
- Phoenix `1.5`
- Absinthe `1.5` 

## 📝 Notes

`erlang-crypto` is not required with the update to Alpine `1.12.0`, so I deployed the app to Heroku to validate `FORCE_SSL` and `DATABASE_SSL` are still fully functional!

https://elixir-boilerplate.herokuapp.com

PR #128, opened 2020-10-14
Update to Elixir 1.11

gcauchon:
## 📖 Description

Following the upgrade I did in #125, here is the conclusion with the update to Elixir `1.11` which includes a minor rename to move from `config/releases.exs` to the now "standard" `config/runtime.exs`.

## 📓 References

An image was deployed Heroku [here](https://elixir-boilerplate.herokuapp.com/)!


Boubalou (approved):
Need to update the notions of versions in README.md and the CHANGELOG.md as well.

.github/workflows/ci.yaml
```diff
@@ -28,6 +28,10 @@ jobs:
     env:
       MIX_ENV: test
       DATABASE_URL: postgres://postgres:boilerplate@localhost/elixir_boilerplate_test
+      PORT: 4001
+      SECRET_KEY_BASE: G0ieeRljoXGzSDPRrYc2q4ADyNHCwxNOkw7YpPNMa+JgP9iGgJKT4K96Bw/Mf/pd
+      SESSION_KEY: test
+      SESSION_SIGNING_SALT: qh+vmMHsOqcjKF3TSSIsghwt2go48m2+IQ+kMTOB3BrSysSr7D4a21uAtt4yp4wn
```
remi, line 34: Is there a way to avoid doing this? Declaring them in `config/test.exs` is so much simpler 😄

test/elixir_boilerplate_web/errors_test.exs
```diff
@@ -63,15 +63,11 @@ defmodule ElixirBoilerplateWeb.ErrorsTest do
...
-                 <li>multiple_roles.type can’t be blank</li>
-                 <li>nicknames should have at least 1 item</li>
-                 <li>single_role.type is invalid</li>
-             </ul>
-           """
+    assert html =~ "<li>email has invalid format</li>"
+    assert html =~ "<li>email should be 10 characters</li>"
+    assert html =~ "<li>multiple_roles.type can’t be blank</li>"
+    assert html =~ "<li>nicknames should have at least 1 item</li>"
+    assert html =~ "<li>single_role.type is invalid</li>"
```
remi, line 70: What is the link between Elixir 1.11 and this? The pull request description doesn’t mention it.

gcauchon: It looks like a collateral from `1.11`…

There are extra `\n` in the generated _safe html_ and modifying the heredoc string to add extra blank lines felt like a patch more than an test intend!

Boubalou: This could be a good candidate to introduce our own HTML testing libs (https://github.com/mirego/html_test_helpers and https://github.com/mirego/html_test_identifiers). These were created for these purpose exactly and would add traction and usage of the libs by the community!

gcauchon: To me, adding libs that are not published to [hex.pm](https://hex.pm/) is like including _beta_ dependencies, I don't see how this would be relevant… 

Especially for the intend of a single test that validates the html serialization of an error changeset!

remi:
@gcauchon Any updates on pending comments?

config/runtime.exs
```diff
@@ -1,47 +1,34 @@
...
-      "1" -> true
-      _ -> false
-    end
-  end
-
-  def get_integer(key) do
-    case get(key) do
-      value when is_bitstring(value) -> String.to_integer(value)
-      _ -> nil
+defmodule RuntimeEnvironment do
```
remi, line 3: Why are there two `Environment` and `RuntimeEnvironment` modules if both of them end up using `System.get_env`?

gcauchon:
> @gcauchon Any updates on pending comments?

Yes! I came up with a neat approach for test environment variables… I'll try to complete the PR today.

.github/workflows/ci.yaml
gcauchon: @remi Could you look at the new approach using `.env.test` and a GitHub Action _step_ to load the dotenv file to `$GITHUB_ENV`

test/elixir_boilerplate_web/errors_test.exs
garno: I’ve added the `hex` badge in both repos to prevent confusion.

* https://github.com/mirego/html_test_identifiers/pull/5
* https://github.com/mirego/html_test_helpers/pull/5

.github/workflows/ci.yaml
remi: Could you explain _why_ we cannot use `config/test.exs` anymore? Is it because it’s not executed _at runtime_ when running `mix test` like before?

gcauchon: It's a matter of compile-time vs. runtime execution in the lifecycle.

> This new configuration file was considered a very useful addition to releases. Therefore, we are also introducing `config/runtime.exs`, which is executed **after** the code compilation on all environments (dev, test, and prod) - for both Mix and releases. Our goal is to provide a better runtime configuration experience to developers, in contrast to our current configuration system which has been mostly compile-time centric.

remi: Right, so we _need_ to have these variables in the environment.

I think your solution works fine! `.env.test` can be used by both CI workflows _and_ by developers  running test in their local enviroment 👍

BOILERPLATE_README.fr.md
```diff
@@ -24,7 +24,7 @@
 
 - Node.js (`>= 12.15, < 13.0`)
 - NPM (`>= 6.13, < 7.0`)
-- Elixir (`~> 1.10`)
+- Elixir (`~> 1.11`)
 - Erlang (`~> 23.1`)
```
remi, line 28: If we have `.tool-versions` now, we don’t need to include them in `BOILERPLATE_README.fr.md` and `BOILERPLATE_README.md` 🎉

gcauchon: 🧹

gcauchon: Should we include a reference to `.tool-versions` instead?

remi: Yes, absolutely! Something like:

```markdown
## 🚧 Dépendances

Toutes les versions des dépendances _runtime_ sont définies dans le fichier `.tool-versions`. Ces dépendances externes sont également requises :

- PostgreSQL (`~> 12.0`)
```

```markdown
## 🚧 Dependencies

All runtime dependency versions are defined in the `.tool-versions` file. These external dependencies are also required:

- PostgreSQL (`~> 12.0`)
```

What do you think?

remi (approved):
🔥

PR #129, opened 2020-10-23
Fix the missing pieces for the Sentry 8.0 and Phoenix 1.5 upgrades

Boubalou:
## 📖 Description

Add the missing pieces from the latest upgrades of libs!

For Sentry, due to the way they expect an HTTP Client by default, they were fall backing to `Hackney`, which is documented here in their 8.0 upgrade documentation: https://gist.github.com/mitchellhenke/dce120a5515565076b13962ee0be749b

Instead of implementing our own behavior corresponding to their expectations, let's define `Hackney` as our go to HTTP Client lib.

This avoid the following start error:

```
** (Mix) Could not start application sentry: exited in: Sentry.start(:normal, [])
    ** (EXIT) an exception was raised:
        ** (RuntimeError) cannot start Sentry.HackneyClient because :hackney is not available.
Please make sure to add hackney as a dependency:
    {:hackney, "~> 1.8"}
            (sentry 8.0.2) lib/sentry/http_client.ex:26: Sentry.HackneyClient.child_spec/0
            (sentry 8.0.2) lib/sentry.ex:101: Sentry.start/2
            (kernel 7.1) application_master.erl:277: :application_master.start_it_old/4
```

### DIFF
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -10219,12 +10219,12 @@
       }
     },
     "phoenix": {
-      "version": "github:phoenixframework/phoenix#b0278829076e30deb994b83bd246966476a9d0b4",
-      "from": "github:phoenixframework/phoenix#v1.4.12"
+      "version": "github:phoenixframework/phoenix#3a5e5c5abf7841700d6ed88c9147f82a55ec2d59",
+      "from": "github:phoenixframework/phoenix#v1.5.5"
     },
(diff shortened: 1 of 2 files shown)


simonprev (approved):

remi (approved):

garno (commented):
Late 👍

PR #131, opened 2020-11-06
Add preferred libraries section in README.md

garno:
## 📖 Description

We had a private list of preferred libraries we use in our projets. There is no reason for this list to be private and can actually help boilerplate users to get started.

These libraries aren’t included by default in the boilerplate since they are not required in all our projects.

✌️

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,22 @@ This boilerplate comes with batteries included, you’ll find:
 4. Create a new Git repository (`git init`)
 5. Create the initial Git commit (`git commit -a -m "Initial commit"`)
 
+## Preferred libraries
+
+Some batteries aren’t included since all projects have their own needs and requirements. Here’s a list of our preferred libraries to help you get started:
+
+| Category                    | Libraries                                                                              |
+| --------------------------- | -------------------------------------------------------------------------------------- |
+| Authentication              | [`ueberauth`](https://hex.pm/packages/ueberauth), [`pow`](https://hex.pm/packages/pow) |
+| Asynchronous job processing | [`oban`](https://hex.pm/packages/oban)                                                 |
+| Emails                      | [`bamboo`](https://hex.pm/packages/bamboo), [`swoosh`](https://hex.pm/packages/swoosh) |
+| File upload                 | [`waffle`](https://hex.pm/packages/waffle)                                             |
+| HTTP client                 | [`tesla`](https://hex.pm/packages/tesla)                                               |
+| HTML parsing                | [`floki`](https://hex.pm/packages/floki)                                               |
+| Pagination                  | [`scrivener`](https://hex.pm/packages/scrivener)                                       |
+| Mocks                       | [`mox`](https://hex.pm/packages/mox), [`mimic`](https://hex.pm/packages/mimic)         |
+| Search                      | [`elasticsearch`](https://hex.pm/packages/elasticsearch)                               |
+
 ## License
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -43,6 +43,22 @@ This boilerplate comes with batteries included, you’ll find:
...
+| Asynchronous job processing | [`oban`](https://github.com/sorentwo/oban)                                                          |
+| Emails                      | [`bamboo`](https://github.com/thoughtbot/bamboo), [`swoosh`](https://github.com/swoosh/swoosh)      |
+| File upload                 | [`waffle`](https://github.com/elixir-waffle/waffle)                                                 |
+| HTTP client                 | [`tesla`](https://github.com/teamon/tesla)                                                          |
+| HTML parsing                | [`floki`](https://github.com/philss/floki)                                                          |
+| Pagination                  | [`scrivener`](https://github.com/drewolson/scrivener)                                               |
+| Mocks                       | [`mox`](https://github.com/dashbitco/mox), [`mimic`](https://github.com/edgurgel/mimic)             |
+| Search                      | [`elasticsearch`](https://github.com/danielberkompas/elasticsearch-elixir)                          |
```
remi, line 60: Shouldn’t we link to Hex.pm packages instead of GitHub projects?

README.md
```diff
@@ -43,6 +43,22 @@ This boilerplate comes with batteries included, you’ll find:
...
+
+Some batteries aren’t included since all projects have their own needs and requirements. Here’s a list of our preferred libraries to help you get started:
+
+| Category                    | Libraries                                                                                           |
+| --------------------------- | --------------------------------------------------------------------------------------------------- |
+| Authentication              | [`ueberauth`](https://github.com/ueberauth/ueberauth), [`pow`](https://github.com/danschultzer/pow) |
+| Asynchronous job processing | [`oban`](https://github.com/sorentwo/oban)                                                          |
+| Emails                      | [`bamboo`](https://github.com/thoughtbot/bamboo), [`swoosh`](https://github.com/swoosh/swoosh)      |
+| File upload                 | [`waffle`](https://github.com/elixir-waffle/waffle)                                                 |
+| HTTP client                 | [`tesla`](https://github.com/teamon/tesla)                                                          |
```
Boubalou, line 56: I believe we enforced Hackney as the default HTTP client in [mix.exs](https://github.com/mirego/elixir-boilerplate/blob/master/mix.exs#L45) due to the latest New Relic dependencies. Not sure if it should removed or if we should change the information here?

garno: `hackney` is required by a lot of libraries, but I still think `tesla` should be recommend for a more high level usage.

README.md
garno: I find the GitHub links more helpful but yeah, I can change it if that’s what we want to expose in the `README.md`.

README.md
remi: Yep, I agree with @garno.

I think Tesla is a much more developer-friendly HTTP client when building business logic code.

I think Hackney is considered more like a low-level HTTP client.

Boubalou: 👍

README.md
Boubalou: I like both. `Hex` is the distribution channel like `NPM` and does expose the `Github` link as well, but you rarely find what you are looking for in the `Hex` website, to be frank.

remi: It’s a small detail indeed, but I think since we’re referencing packages that we include in our `mix.exs`, we should use the "thing that gets pulled from when running mix deps.get" and that’s the hex.pm package 😄

garno:
@remi @Boubalou as soon as this is approved I’ll merge it 😄  I’ve replaced github.com links to hex.pm.

Boubalou (approved):

remi (approved):
👌

PR #132, opened 2020-11-12
Simplification of the Router URLs configuration

Boubalou:
## 📖 Description

Simplification of the Router URLs configuration by reducing the environment variables required to boot up the server! The idea is to reuse through `URI.parse()` the different components instead of asking them separately. 

This change also fixes an issue where the `PORT` was used for the URL when using the Routers generated methods such as `ElixirBoilerplateWeb.Router.Helpers.home_url(ElixirBoilerplateWeb.Endpoint, :index)`.

### DIFF
--- a/config/releases.exs
+++ b/config/releases.exs
@@ -32,26 +32,33 @@ defmodule Environment do
     end
   end
 
-  def get_endpoint_static_url_config(nil), do: nil
-  def get_endpoint_static_url_config(""), do: nil
+  def get_endpoint_url_config(nil), do: nil
+  def get_endpoint_url_config(""), do: nil
 
-  def get_endpoint_static_url_config(host) do
+  def get_endpoint_url_config(uri) do
     [
-      host: host,
-      scheme: get("STATIC_URL_SCHEME"),
-      port: get("STATIC_URL_PORT")
+      host: uri.host,
+      scheme: uri.scheme,
+      port: uri.port
     ]
   end
+
+  def get_uri_part(%URI{host: host}, :host), do: host
+  def get_uri_part(%URI{port: port}, :port), do: port
+  def get_uri_part(%URI{scheme: scheme}, :scheme), do: scheme
+  def get_uri_part(_, _), do: nil
+
+  def get_safe_uri(nil), do: nil
+  def get_safe_uri(""), do: nil
+  def get_safe_uri(url), do: URI.parse(url)
 end
 
-force_ssl = Environment.get_boolean("FORCE_SSL")
-scheme = if force_ssl, do: "https", else: "http"
-host = Environment.get("CANONICAL_HOST")
-port = Environment.get("PORT")
+canonical_uri = Environment.get_safe_uri(Environment.get("CANONICAL_URL"))
+static_uri = Environment.get_safe_uri(Environment.get("STATIC_URL"))
 
 config :elixir_boilerplate,
-  canonical_host: host,
-  force_ssl: force_ssl
+  canonical_host: Environment.get_uri_part(canonical_uri, :host),
+  force_ssl: Environment.get_uri_part(canonical_uri, :scheme) == "https"
 
 config :elixir_boilerplate, ElixirBoilerplate.Repo,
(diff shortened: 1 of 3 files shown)


config/releases.exs
```diff
@@ -32,25 +32,34 @@ defmodule Environment do
...
+      port: uri.port
     ]
   end
+
+  def get_uri_part(nil, _), do: nil
+  def get_uri_part("", _), do: nil
+
+  def get_uri_part(url, part) do
+    url
+    |> URI.parse()
```
remi, line 53: Is there any way we could avoid parsing the URL each time we call this function? Maybe we could parse (safely) both URLs (canonical and static) and _then_ pass those parsed URIs around?

Boubalou: Let me try something, indeed.

Boubalou: Actually, one of the issue here is that URI.parse(nil) isn't supported, therefore requires pattern matching prior to doing it. 🤔

remi: Yeah, we could have a function in `Environment` that parses them safely 😄

Boubalou: I believe I've reached something interesting if you want to take a new look!

config/releases.exs
```diff
@@ -32,25 +32,34 @@ defmodule Environment do
...
-host = Environment.get("CANONICAL_HOST")
-port = Environment.get("PORT")
+canonical_uri = Environment.get_safe_uri(Environment.get("CANONICAL_URL"))
+static_uri = Environment.get_safe_uri(Environment.get("STATIC_URL"))
+force_ssl = Environment.get_uri_part(canonical_uri, :scheme) == "https"
 
 config :elixir_boilerplate,
-  canonical_host: host,
+  canonical_host: Environment.get_uri_part(canonical_uri, :host),
   force_ssl: force_ssl
```
remi, line 63: No need for a separate variable since we now use it only once 🎉 

```suggestion
  force_ssl: Environment.get_uri_part(canonical_uri, :scheme) == "https"
```

Boubalou: Weird, I missed it, lol! Indeed.

config/releases.exs
```diff
@@ -32,26 +32,34 @@ defmodule Environment do
...
-      scheme: get("STATIC_URL_SCHEME"),
-      port: get("STATIC_URL_PORT")
+      host: uri.host,
+      scheme: uri.scheme,
+      port: uri.port
     ]
   end
+
+  def get_uri_part(nil, _), do: nil
+  def get_uri_part("", _), do: nil
```
remi, line 47: Since we always pass a URI or `nil`, I don’t we need this line.

config/releases.exs
```diff
@@ -32,26 +32,34 @@ defmodule Environment do
...
+  def get_uri_part(uri, :port), do: uri.port
+  def get_uri_part(uri, :scheme), do: uri.scheme
+  def get_uri_part(uri, :host), do: uri.host
```
remi, line 50: Would this makes sense?

```suggestion
  def get_uri_part(uri, part), do: Map.get(url, part)
```

remi: In fact we simply could do:

```elixir
def get_uri_part(uri, part) when is_struct(uri, URI), do: Map.get(uri, part)
def get_uri_part(_, _), do: nil
```

or

```elixir
def get_uri_part(%URI{host: host}, :host), do: host
def get_uri_part(%URI{port: port}, :port), do: port
def get_uri_part(%URI{scheme: scheme}, :scheme), do: scheme
def get_uri_part(_, _), do: nil
```

What do you think?

Boubalou: I really like the last approach! I will try it.

remi (approved):
I’d like to see a 👍 from @simonprev before merging this! 😁

Boubalou:
> I’d like to see a 👍 from @simonprev before merging this! 😁

@simonprev, WHERE ARE YOU?

simonprev (approved):
✋ 👍

alarochelle (approved):
🚀

PR #133, opened 2020-11-13
Revert removal of `PORT` environment variable

remi:
## 📖 Description

#132 removed the use of `PORT` in favor of extracting the port from `CANONICAL_URL`.

We still need an explicit `PORT` variable to be able to specify a local port the server will run on vs. the port that will be used to generate URLs.

## 📝 Notes

I also improved the documentation for `CANONICAL_URL` in `.env.dev` to make it clear what’s it’s used for 💪

### DIFF
--- a/.env.dev
+++ b/.env.dev
@@ -11,7 +11,7 @@
 # -----------------------------------------------------------------------------
 
 # Server configuration
-CANONICAL_URL=http://localhost:4000
+PORT=4000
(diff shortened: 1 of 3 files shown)


Boubalou (approved):
Perfect!

simonprev (commented):
👍

charlesdemers (commented):
Late 👍

PR #150, opened 2021-05-29
Erlang 23.3

gcauchon:
## 📖 Description

This the first of a series of updates I would like to propose to the boilerplate based on the recent projects I started with the template.

Before | After
--- | ---
Screen Shot 2021-05-29 at 9 14 25 AM | Screen Shot 2021-05-29 at 9 19 58 AM

## 📝 Notes

I did not update `absinthe`, `new_relic` nor `ex_machina` because I am not using those libs in my projects so I had no significant use cases to validate there were no side effects.

## 🦀 Dispatch

- `#dispatch/elixir`


mix.exs
```diff
@@ -40,16 +40,13 @@ defmodule ElixirBoilerplate.Mixfile do
 
   defp deps do
     [
-      # HTTP Client
-      {:hackney, "~> 1.16"},
```
simonprev, line 44: Why do we remove this deps?

gcauchon: I thought we did not have to explicitly define de dependency, but it looks like `new_relic`needs an explicit definition…

gcauchon:
@remi @simonprev I’d like this PR to be merged so I can complete #152

pretty please

PR #161, opened 2021-11-14
Replace webpack with esbuild

lewazo:
## 📖 Description

This PR replaces webpack by esbuild for assets compiling, which is the default asset bundler shipping with Phoenix v1.6.

## 📝 Notes

We can also remove `phoenix` from the compilers list starting with Phoenix v1.6, which should result in faster compile times.

## 🦀 Dispatch

- `#dispatch/elixir`


Dockerfile
```diff
@@ -76,8 +73,8 @@ COPY priv priv
 RUN mix compile
 
 # Copy assets from step 1
-COPY --from=js-builder /build/priv/static priv/static
-RUN mix phx.digest
+COPY --from=js-builder /build/assets assets
```
gcauchon, line 76: I think we could completely remove the first step of the docker build. Running linters and formaters is a CI concern that could be handled in the GitHub Actions workflow.

```suggestion
COPY assets assets
```

gcauchon (commented):
Any changes required in the `Makefile` targets?

Dockerfile
lewazo: Right, since we removed webpack, only formatting and linting dependencies remains which have no use in the release.

We still, however, need the css reset which is currently added via NPM, so we can't just remove the `npm ci` command.

lewazo:
@gcauchon 
> Any changes required in the Makefile targets?

Nope, looks like it's good to go as-is.

gcauchon (approved):
👌🏻

remi:
Just for this diff:

```diff
-/css/app.css
+/assets/app.css

-/js/app.js
+/assets/app.js
```

I think it’s worth it! 🙌

remi (approved):

garno:
🔥 🚀

PR #163, opened 2021-11-30
Convert CI workflow to “Reusable CI”

remi:
## 📖 Description

This pull request introduces a new GitHub Action workflow: _Reusable CI_. This make it possible for projects to use this simple CI workflow:

```yaml
name: CI

on:
  push:
    branches:
      - master
  pull_request:
    branches:
      - "**"

jobs:
  ci:
    uses: mirego/elixir-boilerplate/.github/workflows/reusable-ci.yaml@master
```

Instead of having to _duplicate_ the whole YAML file inside each project repository.

## 📝 Notes

All references to `feature/convert-ci-to-reusable-ci` will be removed before merging 🤓

## 🦀 Dispatch

- `#dispatch/elixir`

### DIFF
--- a/.github/workflows/reusable-ci.yaml
+++ b/.github/workflows/reusable-ci.yaml
@@ -3,40 +3,39 @@ name: Reusable CI
 on:
   workflow_call:
     inputs:
-      otp-version:
+      database-image:
         required: true
         type: string
       elixir-version:
         required: true
         type: string
+      otp-version:
+        required: true
+        type: string
 jobs:
   reusable-ci:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        otp-version: ${{ inputs.otp-version }}
-        elixir-version: ${{ inputs.elixir-version }}
 
     services:
       db:
-        image: postgres:12
+        image: ${{ inputs.database-image }}
         env:
-          POSTGRES_DB: elixir_boilerplate_test
-          POSTGRES_PASSWORD: boilerplate
+          POSTGRES_DB: database_test
+          POSTGRES_PASSWORD: password
         ports: ["5432:5432"]
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
 
     env:
       MIX_ENV: test
-      DATABASE_URL: postgres://postgres:boilerplate@localhost/elixir_boilerplate_test
+      DATABASE_URL: postgres://postgres:password@localhost/database_test
 
     steps:
       - uses: actions/checkout@v2
 
       - uses: erlef/setup-elixir@v1
         with:
-          elixir-version: ${{ matrix.elixir-version }}
-          otp-version: ${{ matrix.otp-version }}
+          elixir-version: ${{ inputs.elixir-version }}
+          otp-version: ${{ inputs.otp-version }}
 
       - uses: actions/setup-node@v1
         with:
--- a/.github/workflows/reusable-ci.yaml
+++ b/.github/workflows/reusable-ci.yaml
@@ -27,7 +27,6 @@ jobs:
 
     env:
       MIX_ENV: test
-      DATABASE_URL: postgres://postgres:password@localhost/database_test
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/reusable-ci.yaml
+++ b/.github/workflows/reusable-ci.yaml
@@ -1,0 +1,70 @@
+name: Reusable CI
+
+on:
+  workflow_call:
+
+jobs:
+  reusable-ci:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/reusable-ci.yaml
```diff
@@ -0,0 +1,70 @@
...
+        image: postgres:12
+        env:
+          POSTGRES_DB: elixir_boilerplate_test
+          POSTGRES_PASSWORD: boilerplate
+        ports: ["5432:5432"]
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+
+    env:
+      MIX_ENV: test
+      DATABASE_URL: postgres://postgres:boilerplate@localhost/elixir_boilerplate_test
```
remi, line 25: The only difference is that the database used in the CI workflow will be named `elixir_boilerplate_test` (instead of `<project name>_test`). Same thing for the databse user credentials.

I could convert them to [_inputs_](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow) but I think it’s not a big deal, right?

remi:
Also, not a big fan of **CI / ci / reusable-ci** 🤨 I’ll try to come up with something better!

.github/workflows/reusable-ci.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+on:
+  workflow_call:
+
+jobs:
+  reusable-ci:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        otp-version: [24.1.2]
```
JoeyBG, line 11: I think this will be specific to each project! Does this still work if it's an input?

.github/workflows/reusable-ci.yaml
```diff
@@ -0,0 +1,70 @@
...
+        elixir-version: [1.12.3]
+
+    services:
+      db:
+        image: postgres:12
```
JoeyBG, line 16: same thing here

.github/workflows/reusable-ci.yaml
remi: Oh right, that’s a reason to use inputs 🙂

.github/workflows/reusable-ci.yaml
simonprev: We could name it something more generic like `database_test`. If we get an error in our project with the mention of "elixir-boilerplate" it could be confusing.

.github/workflows/reusable-ci.yaml
remi: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onworkflow_callinputsinput_idtype

> Required if input is defined for the on.workflow_call keyword. The value of this parameter is a string specifying the data type of the input. This must be one of: **boolean**, **number**, or **string**.

😩

JoeyBG (commented):
Overall I'm not sure how big of a gain it is to share this kind of workflows across projects.

.github/workflows/reusable-ci.yaml
JoeyBG: I don't think there is a real use case for a matrix for our kind of applications (it's not like we are deploying a library and we want to test for multiple versions at the same time). Maybe we can simplify it a bit?

remi:
> Overall I'm not sure how big of a gain it is to share this kind of workflows across projects.

Isn’t it better than having to duplicate the whole workflow file across project repositories?

.github/workflows/reusable-ci.yaml
remi: Huh, you’re totally right 🤦‍♂️

JoeyBG:
> Isn’t it better than having to duplicate the whole workflow file across project repositories?

Yes but I wonder how often we'll need to modify it and end up duplicating it anyway. For instance, I don't think any of the projects I've worked on recently can use it. As soon as you add an extra step, or add an additional runtime dependency, and you need to copy the whole workflow over and make the required changes.

I also don't think this is the right repository for this kind of shared code across projects.

All things considered, I'm not strongly against it, but I'm not a big fan. I'm not sure the positives outweigh the negatives.

remi:
>> Overall I'm not sure how big of a gain it is to share this kind of workflows across projects.
>
> Isn’t it better than having to duplicate the whole workflow file across project repositories?

A counter-argument is that it would make things a bit more opaque for developers. 😕

remi:
I’m going to close this PR. I wanted to run a quick experiment with reusable workflows and I did 🙂 

But after discussing it with @JoeyBG, we’re still not sure if it’s a good use case for most of our projects, for several reasons (a few of them discussed above ☝️).

PR #164, opened 2022-01-18
Introduce add-translations and order-by flags to accent-sync

pboutin:
## 📖 Description

I suggest adding these 2 options to our default config since they fit most of our projects better.

`--add-translations` since we generally ship bilingual (fr/en) projects for which the developers translates the newly added strings.
`--order-by=key-asc` to help having diffs that are more meaningful after doing a sync (remove all the noise caused by strings that swap places with others).

Thoughts ?

## 🦀 Dispatch

- `#dispatch/elixir`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ dependencies: ## Install dependencies
 
 .PHONY: sync-translations
 sync-translations: ## Synchronize translations with Accent
-	npx accent sync
+	npx accent sync --add-translations --order-by=key-asc
 
 .PHONY: test
 test: ## Run the test suite


remi (approved):
Your explanation makes sense! Any downsides to having these new "defaults"?

gcauchon:
🔠 🏆

simonprev:
> Any downsides to having these new "defaults"

The _only_ downside to `add-translations` is that projets with large number of languages will upload many files on each sync. But since most project have one or two languages I think it is a sensible default value 👍 

And key ordering is good I think, it will make collaboration easier and it does not affect the developer.

simonprev (approved):

lucbernard (approved):

PR #165, opened 2022-02-08
Update hex/npm dependencies

gcauchon:
## 📖 Description

Update `hex` and `npm` dependencies.

The next step will be to review the `esbuild` setup versus a vanilla Phoenix project to ensure the boilerplate did not drift to much apart from `mix phx.new`.

Phoenix `1.7` will be released soon and we want to gap to be as lean as possible.

## 📝 Notes

- Erlang `24.2`
- Elixir `1.13`
- Node `16.13` (LTS)

## 🦀 Dispatch

- `#dispatch/elixir`


assets/package.json
```diff
@@ -4,19 +4,19 @@
...
-    "node": "^14.18",
-    "npm": "^6.14"
+    "node": "^16.13",
+    "npm": "^8.1"
   },
   "dependencies": {
     "simple-css-reset": "^3.0.0"
   },
   "devDependencies": {
     "babel-eslint": "^10.1.0",
```
JoeyBG, line 14: You can switch this to `@babel/eslint-parser` to fix the CI :)

gcauchon: I don’t understand why it is not pulling the dep from the plugin since it’s defined there anyway!?

https://github.com/mirego/eslint-plugin-mirego/blob/master/package.json#L18

JoeyBG: It's listed as a devDependency, I'll go ahead and fix that..

assets/package.json
```diff
@@ -4,19 +4,19 @@
   "private": true,
   "engine-strict": true,
   "engines": {
-    "node": "^14.18",
-    "npm": "^6.14"
+    "node": "^16.13",
+    "npm": "^8.1"
```
Boubalou, line 8: I believe we can skip the npm information here now as it is bundled with Node, if I am not mistaken

gcauchon: We kind of need it for the CI process to run on Github Actions…

bducharme (approved):

PR #166, opened 2022-02-09
Synchronize docker setup with phx.new and phx.gen.release generators

gcauchon:
## 📖 Description

Following the dependencies update in #165, here are a few changes the release/docker setup to follow the community setup introduce in `1.6` and generated by `phx.gen.release`.

## 📝 Notes

Here are the important parts:

1. In a vanilla setup, the initial `npm-builder` step could be avoided, but since we are still depending on `simple-css-reset` we need it to install it first for `esbuild` to be able to bundle it. In an ideal world, the reset is handled by the CSS framework, not an explicit dependency.

2. The "recommended" base image is Debian, not Alpine to prevent DNS resolution issue. See the comment in the generator [here](https://github.com/phoenixframework/phoenix/blob/master/priv/templates/phx.gen.release/Dockerfile.eex#L1-L2).

3. Instead of using a docker `ENTRYPOINT`, `phx.gen.release` generates shell scripts in `bin`, next to the release binary. This implies that migrations have to be explicitly instead of automatically ran each the release boots up. This might be something we want to address differently…

## 🦀 Dispatch

- `#dispatch/elixir`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,8 +13,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        otp-version: [24.2.1]
-        elixir-version: [1.13.2]
+        otp-version: [24.3]
+        elixir-version: [1.13]
 
     services:
       db:
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
-erlang 24.2.1
-elixir 1.13.2-otp-24
-nodejs 16.13.2
+erlang 24.3.3
+elixir 1.13.3-otp-24
+nodejs 16.14.2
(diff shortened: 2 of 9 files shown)


simonprev:
Does it runs the migration automatically before starting the app?

gcauchon:
> Does it runs the migration automatically before starting the app?

@simonprev No! And I think it actually is a good thing… We had a side discussion with @remi and @JoeyBG about moving the migration "trigger" to the deployment phase would make more sense. What is your take on this matter?

gcauchon:
@simonprev I restored migrations in the startup script like we use to do with `docker-entrypoint.sh`!

gcauchon:
I am merging PRs today!

I'll do a dryrun of the setup to fix potential issues…

PR #168, opened 2022-03-22
Add docker buildx in CI workflow to improve docker build speed

simonprev:
## 📖 Description

https://docs.docker.com/buildx/working-with-buildx

In one of our project, it reduced the Docker build time from 7 minutes to 5 minutes 🐎 💨 

## 🦀 Dispatch

- `#dispatch/elixir`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -72,4 +72,11 @@ jobs:
       - run: make check
       - run: make lint
       - run: make test
+
+      - name: Install Docker Buildx
+      - uses: docker/setup-buildx-action@v1
+        id: docker-buildx
+        with:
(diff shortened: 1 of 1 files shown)


gcauchon:
Do we want to make a similar change to the `Makefile`? This way we could leverage the efficiency gain when building/testing the image locally too…

simonprev:
> This way we could leverage the efficiency gain when building/testing the image locally too…

Since docker buildx, when installed, replace the `docker build` command, I think we should leave it to the developer to install such tool

remi (approved):

gcauchon (approved):

marcantoinefortier (approved):

PR #172, opened 2022-04-21
Extract the helpers module out of runtime.exs

gcauchon:
## 📖 Description

I was reading [this blog post](https://felt.com/blog/elixir-configuration) by @axelson yesterday and I really liked the approach proposed in point #2, especially the reference to the `ConfigHelpers` module from the Code::Stats repository!

> 2. **Use configuration helpers.** I like to use configuration helpers which are inspired by this [file](https://gitlab.com/code-stats/code-stats/-/blob/b1cf53462a3fa34369eaa06494754c7ae38aed2a/lib/code_stats/config_helpers.ex). Using configuration helpers like this eases the burden of converting string environment variables into their proper form which may be a boolean or an integer. Once I have the config helpers imported I can read an environment variable with `get_env("RUN_SEEDS_ON_STARTUP", false, :boolean)` where false is the default if the environment variable `RUN_SEEDS_ON_STARTUP` is not provided and `:boolean` means to parse the environment variable into either false or true (any values that can’t be parsed to one of those is treated as an error). 

It’s a similar approach, but it simplifies `runtime.exs` to makes it significantly easier to read and understand.

## 📓 References

- [Tips for Improving Your Elixir Configuration](https://felt.com/blog/elixir-configuration)
- `ConfigHelpers` module from the [`Code::Stats` repository](https://gitlab.com/code-stats/code-stats/-/blob/b1cf53462a3fa34369eaa06494754c7ae38aed2a/lib/code_stats/config_helpers.ex)

## 🦀 Dispatch

- `#dispatch/elixir`

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -1,99 +1,42 @@
 import Config
+import ElixirBoilerplate.ConfigHelpers
 
-defmodule Environment do
-  @moduledoc """
-  This modules provides various helpers to handle environment metadata
-  """
-
-  def get(key), do: System.get_env(key)
-
-  def get_boolean(key) do
-    case get(key) do
-      "true" -> true
-      "1" -> true
-      _ -> false
-    end
-  end
-
-  def get_integer(key) do
-    case get(key) do
-      value when is_bitstring(value) -> String.to_integer(value)
-      _ -> nil
-    end
-  end
-
-  def get_cors_origins do
-    case Environment.get("CORS_ALLOWED_ORIGINS") do
-      origins when is_bitstring(origins) ->
-        origins
-        |> String.split(",")
-        |> case do
-          [origin] -> origin
-          origins -> origins
-        end
-
-      _ ->
-        nil
-    end
-  end
-
-  def get_endpoint_url_config(nil), do: nil
-  def get_endpoint_url_config(""), do: nil
-
-  def get_endpoint_url_config(uri) do
-    [
-      host: uri.host,
-      scheme: uri.scheme,
-      port: uri.port
-    ]
-  end
-
-  def get_uri_part(%URI{host: host}, :host), do: host
-  def get_uri_part(%URI{port: port}, :port), do: port
-  def get_uri_part(%URI{scheme: scheme}, :scheme), do: scheme
-  def get_uri_part(_, _), do: nil
-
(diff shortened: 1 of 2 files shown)


pboutin (approved):

lib/elixir_boilerplate/config_helpers.ex
```diff
@@ -0,0 +1,53 @@
+# credo:disable-for-this-file CredoNaming.Check.Warning.AvoidSpecificTermsInModuleNames
+defmodule ElixirBoilerplate.ConfigHelpers do
```
remi, line 2: Since they are all environment helpers, should we just rename it to

```suggestion
defmodule ElixirBoilerplate.Environment do
```

Or something like that?

remi (commented):
~`Successfull` → `Successful` 🤓~

Wrong PR! 🤦‍♂️

lib/elixir_boilerplate/config_helpers.ex
gcauchon: @remi At the beginning, I extracted the module and kept the `Environment` name, but it felt weird to import functions from a module with a name too similar to `System` and `Application` from the stdlib.

Like the comment state, it's a set of helper functions for Elixir [`Config`](https://hexdocs.pm/elixir/Config.html), not an `Environment` set of function 🤔

remi: Alright, how about

```suggestion
defmodule ElixirBoilerplate.Config do
```

then?

remi: Or since they are helpers used in a single file, why extract them into an external module at all?

gcauchon: The cognitive load of defining a module inside the (`runtime.exs`) script always felt weird imho. It pushes `runtime.exs` in a way that differs greatly from the other `Config` scripts and breaks it apart from the language conventions.

remi: And how about my previous comment? 🙂

gcauchon: It's perfect, already made the change! Just haven't had time to push it yet…

remi (approved):

PR #173, opened 2022-06-14
Erlang 25.0

gcauchon:
## 📖 Description

Since the `JIT` is now enabled on ARM-based computer following the release of Erlang `25`, let's upgrade the boilerplate to use the latest the latest versions as a new baseline 🎉 

## 📓 References

- https://www.erlang.org/blog/my-otp-25-highlights/#improvements-of-the-jit

> The JIT now supports the [AArch64 (ARM64)](https://en.wikipedia.org/wiki/AArch64) architecture, used by (for example) [Apple Silicon](https://en.wikipedia.org/wiki/Apple_silicon) Macs and newer [Raspberry Pi](https://en.wikipedia.org/wiki/Raspberry_Pi) devices.

## 🦀 Dispatch

- `#dispatch/elixir`


.tool-versions
```diff
@@ -1,3 +1,3 @@
-erlang 24.2.1
-elixir 1.13.2-otp-24
-nodejs 16.13.2
+erlang 25.0.1
+elixir 1.13.4-otp-25
+nodejs 16.15.0
```
Boubalou, line 3: Latest LTS c'est `16.15.1`

Boubalou (approved):

.tool-versions
gcauchon: `^TRUE`, I'll fix it!

PR #174, opened 2022-07-28
Add priv/null as source and custom Interpolation module for gettext in test env

simonprev:
## 📖 Description

The goal is to have the key as output in templates. This way, we can test the content without relying on `po` files.
With the custom Interpolation module, bindings are also printed in the templates

```elixir
Gettext.dgettext(ElixirBoilerplate.Gettext, "errors", "test", %{arg: "ok"})
#=> test[arg=ok]
```

## 🦀 Dispatch

- `#dispatch/elixir`

### DIFF
--- a/test/support/gettext_interpolation.ex
+++ b/test/support/gettext_interpolation.ex
@@ -1,0 +1,47 @@
+defmodule ElixirBoilerplate.GettextInterpolation do
+  @moduledoc """
+  Default Gettext.Interpolation implementation for testing purposes
+
+  Appends formatted `bindings` at the end of the string
+  """
+
+  @behaviour Gettext.Interpolation
+
+  @doc """
+  iex> runtime_interpolate("test", %{})
+  {:ok, "test"}
+  iex> runtime_interpolate("test", %{arg: 1})
+  {:ok, "test[arg=1]"}
+  iex> runtime_interpolate("test", %{arg: :atom})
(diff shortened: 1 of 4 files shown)


remi:
It works so well it makes the right test fails! 😁

gcauchon (approved):

MathieuLegault1 (approved):

remi (approved):
💥

remi:
@simonprev Could you update the PR title now that the content is much more awesome? 😀

test/support/gettext_interpolation.ex
```diff
@@ -0,0 +1,47 @@
...
+  defp format_bindings(bindings) when is_map(bindings) and map_size(bindings) === 0, do: ""
+
+  defp format_bindings(bindings) when is_map(bindings) do
+    bindings = Enum.map_join(bindings, ",", fn {key, value} -> "#{key}=#{format_value(value)}" end)
+    "[#{bindings}]"
+  end
+
+  defp format_bindings(_bindings), do: ""
+
+  defp format_value(value) when is_list(value), do: Enum.map_join(value, ",", &format_value/1)
```
gcauchon, line 44: Should we wrap the list in square brackets here too?

simonprev: I think we can represent list of value with simple comma separated values, the goal is to allow matching of strings in test cases so the simpler the better

PR #178, opened 2022-08-09
Include Dialyzer

gcauchon:
## 📖 Description

`Dialyzer` was removed in #61, before we had enough time to explore and adapt our workflows and pipelines to use the full power of static type analysis.

We are using it in many projects now and it’s good practice to start fro this template with static analysis enabled! 

## 🦀 Dispatch

- `#dispatch/elixir`


.github/workflows/ci.yaml
```diff
@@ -39,29 +39,22 @@ jobs:
...
-      - uses: actions/cache@v2
-        id: build-cache
-        with:
-          path: _build
-          key: ${{ runner.os }}-build-${{ matrix.otp-version }}-${{ matrix.elixir-version }}-${{ hashFiles(format('{0}/mix.lock', github.workspace)) }}
+          path: |
+            deps
+            _build
+            priv/plts
+          key: ${{ runner.os }}-mix-${{ matrix.otp-version }}-${{ matrix.elixir-version }}-${{ hashFiles(format('{0}/mix.lock', github.workspace)) }}
```
gcauchon, line 51: Since I added the _Persistent Lookup Table_ folder to optimize the Dialyzer workload, I merged the `mix`-related caching steps into a single hash key so everything related to `mix` is cached altogether.

.github/workflows/ci.yaml
```diff
@@ -39,29 +39,22 @@ jobs:
...
 
       - uses: actions/cache@v2
         id: npm-cache
         with:
           path: ~/.npm
           key: ${{ runner.os }}-npm-${{ hashFiles(format('{0}/assets/package-lock.json', github.workspace)) }}
-          restore-keys: |
```
simonprev, line 63: This is not valid?

gcauchon: If I remember right, this was not useful… I'll have to go back to the documentation!

gcauchon: > **Optional** A string containing alternative restore keys, with each restore key placed on a new line. If no cache hit occurs for key, these restore `keys` are used sequentially in the order provided to find and restore a cache. 

https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#input-parameters-for-the-cache-action

.github/workflows/ci.yaml
```diff
@@ -39,29 +39,22 @@ jobs:
 
       - uses: actions/setup-node@v1
         with:
```
JoeyBG, line 41: You could replace this with 

```
      - uses: actions/setup-node@v3
         with:
	    node-version: 16.15
	    cache: 'npm'
```

And remove npm cache

.github/workflows/ci.yaml
```diff
@@ -13,12 +13,12 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
```
JoeyBG, line 14: I'm a fan of just removing this and inlining the versions below, since we don't actually maintain the boilerplate (or any application) for multiple elixir/erlang versions

JoeyBG (approved):

.github/workflows/ci.yaml
remi: Oh right, I don’t know why we’re using a matrix strategy here 🤔

gcauchon: Will do!

gcauchon: @JoeyBG However, the `matrix` is helpful the cache key interpolation. Is there any "variables" concepts available in _GitHub Actions_?

remi: Yes, I think you’ll have to use [environment variables](https://docs.github.com/en/actions/learn-github-actions/environment-variables).

gcauchon: > Yes, I think you’ll have to use [environment variables](https://docs.github.com/en/actions/learn-github-actions/environment-variables).

I also saw this in the documentation, but was not sure if it made sense to uses `ENV` for this purpose 🤔

.github/workflows/ci.yaml
```diff
@@ -11,57 +11,42 @@ on:
...
 
-      - uses: actions/setup-node@v1
+      - uses: actions/cache@v3
         with:
-          node-version: 16.13
+          path: |
+            deps
+            _build
+            priv/plts
+          key: ${{ runner.os }}-mix-${{ env.ERLANG_VERSION }}-${{ env.ELIXIR_VERSION }}-${{ hashFiles(format('{0}/mix.lock', github.workspace)) }}
```
gcauchon, line 44: @remi @JoeyBG what do you guys think about this? I can not say it's simpler/cleaner this way 🤔

gcauchon, line 44: @remi @JoeyBG what do you guys think about this? I can not say it's simpler/cleaner this way 🤔

JoeyBG: No, it's not great indeed. I didn't think about the cache..

JoeyBG: Maybe we leave this _as-is_ and push a PR upstream to include the cache automatically in the future (like setup-node, setup-ruby and a few others support).

gcauchon: This is what I was about to propose!

Makefile
```diff
@@ -138,3 +138,10 @@ lint-scripts:
 .PHONY: lint-styles
 lint-styles:
 	cd assets && npx stylelint $(STYLES_PATTERN)
+
+# Static code analysis
+# ----------------------------
+
+.PHONY: static-analysis
+static-analysis: ## Run static code analysis
```
remi, line 146: Should we just create a new `check-` target? This is how most of the other tools are used.

Something like `make check-static-analysis`.

gcauchon: Yes, I found `check` to be a synonym to `analisys`, but consistency always wins!

`check-static-typing` might be better?

Makefile
```diff
@@ -117,14 +117,18 @@ check-format:
 check-unused-dependencies:
 	mix deps.unlock --check-unused
 
+.PHONY: check-static-typing
+check-static-typing:
+	mix dialyzer
+
```
gcauchon, line 123: @remi what do you think about this? I also include it in the `check` target.

remi: Yup! 👌

mix.exs
```diff
@@ -5,8 +5,8 @@ defmodule ElixirBoilerplate.Mixfile do
     [
       app: :elixir_boilerplate,
       version: "0.0.1",
-      elixir: "~> 1.12",
-      erlang: "~> 24.1",
+      erlang: "~> 25.0",
+      elixir: "~> 1.13",
```
remi, line 9: Is this also related to Dialyzer?

gcauchon: It's a leftover from #166, it looks like I forgot to upgrade the specs in `mix.exs`

lib/elixir_boilerplate_graphql/router.ex
```diff
@@ -10,7 +10,6 @@ defmodule ElixirBoilerplateGraphQL.Router do
     forward("/",
       to: Absinthe.Plug,
       init_opts: [
-        document_providers: {ElixirBoilerplateGraphQL, :document_providers},
```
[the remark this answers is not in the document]
gcauchon: Doing so would actually break CI, as I explained in the PR description. I can do it this way too…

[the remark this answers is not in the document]
gcauchon: @garno will introduce APR in a future PR when the setup will be production ready based on the client project he is working on right now.

PR #180, opened 2022-08-11
Remove explicit newrelic configuration

JoeyBG:
## 📖 Description

This is the default behaviour of the agent, no need to explicitly load those variables!

## 🦀 Dispatch

- `#dispatch/elixir`


garno (approved):

remi:
I’m wondering if it’s a _good_ thing to be explicit instead of implicit here 🤔

JoeyBG:
> I’m wondering if it’s a _good_ thing to be explicit instead of implicit here 🤔

I wasn't sure either. The variable is still documented in `.env.dev` (and in the official [documentation](https://hexdocs.pm/new_relic_agent/readme.html#configuration))

PR #230, opened 2023-01-06
Use own one-liner script in CI workflow to inject environment variables from .env.test

remi:
## 📖 Description

Let’s try to use less third-party actions for simple things we can do ourselves 🙂

## 📝 Notes

This is my first try, hopefully it’ll work! ✌️

## 🦀 Dispatch

- `#dispatch/elixir`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -50,10 +50,7 @@ jobs:
           node-version-file: .tool-versions
           cache: npm
 
-      - name: Setup environment variables
-        uses: c-py/action-dotenv-to-setenv@v2
-        with:
-          env-file: .env.test
+      - run: grep -v '^\(#.*\|\s\?\)$' .env.test >> $GITHUB_ENV
 
       - run: make prepare
       - run: make check


.github/workflows/ci.yaml
```diff
@@ -54,10 +53,7 @@ jobs:
           node-version: 16.15
           cache: 'npm'
 
-      - name: Setup environment variables
-        uses: c-py/action-dotenv-to-setenv@v2
-        with:
-          env-file: .env.test
+      - run: grep -v '^\(#.*\|\s\?\)$' .env.test >> $GITHUB_ENV
```
JoeyBG, line 56: 😨

JoeyBG (approved):

.github/workflows/ci.yaml
JoeyBG: > 10 years coding and I still can't read RegEx

gcauchon (approved):

.github/workflows/ci.yaml
Boubalou: same!

PR #241, opened 2023-01-26
Add Plugs.Security in main endpoint, not individual routers

remi:
## 📖 Description

We had a module (`FooWeb.ContentSecurityPolicy`) to provide secure HTTP headers (including a custom `Content-Security-Policy`) that we were injecting in our main router.

But what about our Health and GraphQL routers? 🙅‍♂️

Not anymore! We now have a `FooWeb.Plugs.Security` module (that does the same thing as the existing one) that we inject in `FooWeb.Endpoint` instead. Problem solved!

## 🦀 Dispatch

- `#dispatch/elixir`

### DIFF
--- a/lib/elixir_boilerplate_web/endpoint.ex
+++ b/lib/elixir_boilerplate_web/endpoint.ex
@@ -49,6 +49,7 @@ defmodule ElixirBoilerplateWeb.Endpoint do
   plug(Plug.MethodOverride)
   plug(Plug.Head)
 
+  plug(ElixirBoilerplateWeb.Plugs.Security)
   plug(ElixirBoilerplateHealth.Router)
   plug(ElixirBoilerplateGraphQL.Router)
   plug(:halt_if_sent)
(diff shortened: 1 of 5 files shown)


gcauchon:
A follow up reflection could be: does it still make sense to have separate `router`? Especially considering the _workaround_ we had to put in place to halt the `Plug` pipeline in the `Endpoint`.

lib/elixir_boilerplate_web/endpoint.ex
```diff
@@ -49,6 +49,7 @@ defmodule ElixirBoilerplateWeb.Endpoint do
   plug(Plug.MethodOverride)
   plug(Plug.Head)
 
+  plug(ElixirBoilerplateWeb.Plugs.Security)
   plug(ElixirBoilerplateHealth.Router)
   plug(ElixirBoilerplateGraphQL.Router)
   plug(:halt_if_sent)
```
gcauchon, line 55: Here!

simonprev (approved):

remi:
I’m going to merge this.

@gcauchon Sure, please send a new pull request if you have potential changes you want to discuss! 💪

PR #245, opened 2023-02-06
Disable GraphQL introspection by default

remi:
## 📖 Description

We want to [disable introspection](https://www.apollographql.com/blog/graphql/security/why-you-should-disable-graphql-introspection-in-production/) by default for GraphQL APIs.

## 📝 Notes

Absinthe requires modifying the [pipeline](https://hexdocs.pm/absinthe/Absinthe.Pipeline.html) at compile-time, so a runtime environment variable is out of the question. Thanks @JoeyBG for suggesting to use `Application.compile_env` and hardcoded boolean values in config files! 🎩 

## 🦀 Dispatch

- `#dispatch/elixir`

### DIFF
--- a/lib/elixir_boilerplate_graphql/schema.ex
+++ b/lib/elixir_boilerplate_graphql/schema.ex
@@ -3,6 +3,20 @@ defmodule ElixirBoilerplateGraphQL.Schema do
 
   alias ElixirBoilerplate.Repo
 
+  defmodule Introspection do
+    @disable_introspection Application.compile_env(:elixir_boilerplate, ElixirBoilerplateGraphQL)[:enable_introspection]
+
+    def pipeline(pipeline) do
+      if @disable_introspection do
+        pipeline
+      else
+        Absinthe.Pipeline.without(pipeline, Absinthe.Phase.Schema.Introspection)
+      end
+    end
+  end
+
+  @pipeline_modifier Introspection
+
   import_types(Absinthe.Type.Custom)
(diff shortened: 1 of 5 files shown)


lib/elixir_boilerplate_graphql/schema.ex
```diff
@@ -3,6 +3,20 @@ defmodule ElixirBoilerplateGraphQL.Schema do
...
+    @disable_introspection Application.compile_env(:elixir_boilerplate, ElixirBoilerplateGraphQL)[:enable_introspection]
+
+    def pipeline(pipeline) do
+      if @disable_introspection do
+        pipeline
+      else
+        Absinthe.Pipeline.without(pipeline, Absinthe.Phase.Schema.Introspection)
+      end
+    end
+  end
```
simonprev, line 16: Since it’s a compile time module attributes, we can save the conditional at runtime. Also I think the condition is not correct?

```suggestion

    if @disable_introspection do
       def pipeline(pipeline), do: Absinthe.Pipeline.without(pipeline, Absinthe.Phase.Schema.Introspection)
    else
      def pipeline(pipeline), do: pipeline
    end
```

remi: Oh, good catch! The condition was right, but the naming was wrong indeed…

```elixir
@disable_introspection Application.compile_env(…)[:enable_introspection]
```

🤦‍♂️

JoeyBG (approved):

PR #249, opened 2023-02-08
WIP — Realign boilerplate with `mix phx.new`

gcauchon:
## 📖 Description

Now that the official release of Phoenix `1.7` have seen multiple maintenance releases, here is an open discussion about how we could adapt our _boilerplate_ project to follow the community conventions to simplify future evolution as things are moving forward rapidly in the Erlang/Elixir/Phoenix ecosystem.

## 📓 References

I used the community tool to highlight the gap between `1.6` and `1.7`:
- https://www.phoenixdiff.org/?source=1.6.15&source_variant=default&target=1.7.6&target_variant=default

I had to change a few things because we diverged from the community patterns a few version back and it doesn't fit well in the new _functional component_ approach. Especially if we want to leverage component reusability between _dead_ and _live_ views:

- [x] Added *Tailwind* alongside *Esbuild*;
- [x] Completed the migration to `heex` with the new "view-module-less" approach;
	- I had to disable `CredoNaming.Check.Consistency.ModuleFilename` for now since the validation rule do not match the new file structure convention with functional components.
- [x] I merged every routers back to a single `ElixirBoilerplateWeb.Router` module. 
	- Using multiples `router` and a custom `plug`  to catch an halted `Conn` was a red flag in my mind;
- [x] Re-introduced the `ElixirBoilerplateWeb` module and the `__using__` macro to encapsulate the router/controller/html/live_view and live_component boilerplate like `phx.new` and the generators does;
- [x] Updated the `Absinthe` configuration and added the missing dependency to support `Subscription` over web socket;
- [x] Introduce a `Session` module which pulls the session `key` and `signing_salf` from environment variable at runtime. So we can use it right in both _dead_ and _live_ views;
- [x] Added a default `Ecto` schema and it corresponding migration so we are using `--binary_id`.

## 🦀 Dispatch

- `#dispatch/elixir`


remi:
> ⚠️ I am breaking up https://github.com/mirego/elixir-boilerplate/pull/249 in smaller chunks to simplify open discussion on the many aspect where our boilerplate project diverted from phx.new to realign our use of Phoenix, Absinthe and many libraries with the ongoing effort of the Elixir community.

@gcauchon Since you mentioned this ☝️ in #359, we can close this PR? (Or at least put it back in draft mode)

gcauchon:
Yes!

PR #250, opened 2023-02-13
Add support for ECTO_IPV6 env var for repo socket options

simonprev:
## 📖 Description

This has been merged in `phoenix` for a long time. There is no drawback in our setup, just less headaches if we want to deploy our app in a different provider like fly.io.

## 🦀 Dispatch

- `#dispatch/elixir`

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -11,7 +11,8 @@ config :elixir_boilerplate,
 config :elixir_boilerplate, ElixirBoilerplate.Repo,
   url: get_env!("DATABASE_URL"),
   ssl: get_env("DATABASE_SSL", :boolean),
-  pool_size: get_env!("DATABASE_POOL_SIZE", :integer)
+  pool_size: get_env!("DATABASE_POOL_SIZE", :integer),
+  socket_options: if(get_env("ECTO_IPV6", :boolean), do: [:inet6], else: [])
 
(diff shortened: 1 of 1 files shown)


remi (approved):

config/runtime.exs
```diff
@@ -11,7 +11,8 @@ config :elixir_boilerplate,
 config :elixir_boilerplate, ElixirBoilerplate.Repo,
   url: get_env!("DATABASE_URL"),
   ssl: get_env("DATABASE_SSL", :boolean),
-  pool_size: get_env!("DATABASE_POOL_SIZE", :integer)
+  pool_size: get_env!("DATABASE_POOL_SIZE", :integer),
+  socket_options: if(get_env("ECTO_IPV6", :boolean), do: [:inet6], else: [])
```
gcauchon, line 15: Souldn't we use a name aligned with other `DATABASE_` environment variables?

Something like `DATABASE_IPV6` might make more sense along side the multiple ENV we already have.

gcauchon: There are not many `ENV` in a vanilla `phx.new` project and we added a few important pieces to get our baseline as _12-Factor_ as possible.

We might want to add it in `.env.dev` and `.env.test` (even if there’s a smart default in `ElixirBoilerplate.Config.get_env/2` ) for documentation/discoverability.

PR #258, opened 2023-02-14
Add minimal LiveView setup with a /live example page

simonprev:
## 📖 Description

This is an attempt at a minimal setup for LiveView inspired by an internal project at Mirego. *We must wait for the phoenix 1.7 release to drop before merging this*.

- Shared templates between live and dead components (`/` and `/live`)
- No more `Phoenix.View`
- Minimal `HTML` module that maps to the new HTML/JSON module in phoenix to use in controllers
- Only the unified `Phoenix.Flash` integration needs phoenix 1.7.

image

The goal is to have this exact setup in a deployed project before merging it, but since the changes are not that big, it will serve as a reference for when phoenix 1.7 drops.

## 🦀 Dispatch

- `#dispatch/elixir`


lib/elixir_boilerplate_web/session.ex
```diff
@@ -0,0 +1,9 @@
+defmodule ElixirBoilerplateWeb.Session do
+  def config do
+    [
+      store: :cookie,
+      key: Application.get_env(:elixir_boilerplate, ElixirBoilerplateWeb.Endpoint)[:session_key],
+      signing_salt: Application.get_env(:elixir_boilerplate, ElixirBoilerplateWeb.Endpoint)[:session_signing_salt]
+    ]
+  end
```
gcauchon, line 8:
```suggestion
  def config do
    [
      store: :cookie,
      key: app_config(:session_key),
      signing_salt: app_config(:session_signing_salt)
    ]
  end

  defp app_config(key), do: Application.get_env(:elixir_boilerplate, ElixirBoilerplateWeb.Endpoint)[key]
```

lib/elixir_boilerplate_web/layouts/templates/flash.html.heex
```diff
@@ -0,0 +1,10 @@
+<% id = "flash-" <> to_string(@kind) %>
```
gcauchon, line 1: I was under the impression defining variable match inside of heex template was breaking the static/dynamic checks in the live view stateful process?

mix.exs
```diff
@@ -57,8 +57,9 @@ defmodule ElixirBoilerplate.Mixfile do
       {:corsica, "~> 1.3"},
 
       # Phoenix
-      {:phoenix, "~> 1.6"},
+      {:phoenix, "~> 1.7.0-rc.2", override: true},
```
gcauchon, line 60: Since we are upgrading to `1.7` why don’t we leverage _functional components_ more instead of collocating heex templates?

We could also use the `attr` and `slot` definition macros at the sametime to gain compile time type checking inside of the html/component semantic.

assets/js/app.js
```diff
@@ -1,2 +1,34 @@
...
+import {Socket} from 'phoenix';
+import {LiveSocket} from 'phoenix_live_view';
+
+const FLASH_TTL = 8000;
+const Hooks = {};
+
+Hooks.Flash = {
+  mounted() {
+    const hide = () =>
+      liveSocket.execJS(this.el, this.el.getAttribute('phx-click'));
```
gcauchon, line 13:
```suggestion
    const hide = () => {
      liveSocket.execJS(this.el, this.el.getAttribute('phx-click'));
    };
```

assets/package.json
```diff
@@ -8,6 +8,9 @@
     "npm": "^9.5.1"
   },
   "dependencies": {
+    "phoenix": "^1.7.7",
+    "phoenix_html": "^3.3.1",
+    "phoenix_live_view": "^0.19.5",
```
gcauchon, line 13: If I remember right, with `esbuild`, we do not have to explicitly maintain JS dependencies.

simonprev: We could use the Javascript deps inside the `deps` folder built-in with the elixir packages but this means we have to change how we build our Docker application. This could be done in another pull request!

gcauchon (approved):

PR #267, opened 2023-03-03
Replace backticks with “shell” make utility

remi:
## 📖 Description

Using make’s [shell utility](https://www.gnu.org/software/make/manual/html_node/Shell-Function.html) is the right way (instead of backticks) to execute runtime commands and use outputs in targets.

## 📝 Notes

🎩 Hat tip to @Boubalou for this piece of knowledge 😄 

## 🦀 Dispatch

- `#dispatch/devops`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,13 @@
 # Build configuration
 # -------------------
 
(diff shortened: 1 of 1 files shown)


JoeyBG (approved):
TIL!

Boubalou (approved):
👌

PR #274, opened 2023-03-17
Fix typo in GraphQL schema comment and in .env.test

simonprev:
## 📖 Description

- `.env.test` uses the same database as the `.env.dev.local` in the repo. The password should be the same in both files.
- Typo in schema comment

## 🦀 Dispatch

- `#dispatch/elixir`

### DIFF
--- a/lib/elixir_boilerplate_graphql/schema.ex
+++ b/lib/elixir_boilerplate_graphql/schema.ex
@@ -10,9 +10,8 @@ defmodule ElixirBoilerplateGraphQL.Schema do
     import_fields(:application_queries)
   end
 
-  # Even if having an empty mutation block is valid and works in Absinthe, it
-  # causes a JavaScript error in GraphiQL so uncomment it when you add the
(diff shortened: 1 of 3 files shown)


lib/elixir_boilerplate_graphql/schema.ex
```diff
@@ -10,8 +10,8 @@ defmodule ElixirBoilerplateGraphQL.Schema do
     import_fields(:application_queries)
   end
 
-  # Even if having an empty mutation block is valid and works in Ansinthe, it
-  # causes a Javascript error in GraphiQL so uncomment it when you add the
```
JoeyBG, line 14: Is GraphiQL still included in the boilerplate?

simonprev: No, but still, we don’t want to generate an invalid schema for external tools that could use our GraphQL endpoint.

gcauchon (approved):

PR #359, opened 2023-07-31
Use UUIDs as primary keys instead of sequential integers

gcauchon:
## 📖 Description

⚠️ I am breaking up #249 in smaller chunks to simplify open discussion on the many aspect where our boilerplate project diverted from `phx.new` to realign our use of Phoenix, Absinthe and many libraries with the ongoing effort of the Elixir community.

We are always using `UUID` as primary keys, but since the boilerplate does not inclure any Ecto schema, the groundwork was never migrated back to the template project.

## 📝 Notes

- `use ElixirBoilerplate.Schema` instead of `use Ecto.Schema` in every schema module.
- The new settings on `ElixirBoilerplate.Repo` sets the key type for migrations.

## 🦀 Dispatch

- `#dispatch/elixir`

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -14,7 +14,7 @@ config :elixir_boilerplate, ElixirBoilerplateWeb.Endpoint,
 
 config :elixir_boilerplate, ElixirBoilerplate.Repo,
   migration_primary_key: [type: :binary_id, default: {:fragment, "gen_random_uuid()"}],
-  migration_timestamps: [type: :utc_datetime],
+  migration_timestamps: [type: :utc_datetime_usec],
   start_apps_before_migration: [:ssl]
 
 config :elixir_boilerplate, Corsica, allow_headers: :all
--- a/config/config.exs
+++ b/config/config.exs
@@ -13,7 +13,8 @@ config :elixir_boilerplate, ElixirBoilerplateWeb.Endpoint,
   render_errors: [view: ElixirBoilerplateWeb.Errors.View, accepts: ~w(html json)]
 
 config :elixir_boilerplate, ElixirBoilerplate.Repo,
-  migration_primary_key: [name: :id, type: :binary_id],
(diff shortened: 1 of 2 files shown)


config/config.exs
```diff
@@ -12,7 +12,9 @@ config :elixir_boilerplate, ElixirBoilerplateWeb.Endpoint,
   pubsub_server: ElixirBoilerplate.PubSub,
   render_errors: [view: ElixirBoilerplateWeb.Errors.View, accepts: ~w(html json)]
 
-config :elixir_boilerplate, ElixirBoilerplate.Repo, start_apps_before_migration: [:ssl]
+config :elixir_boilerplate, ElixirBoilerplate.Repo,
+  migration_primary_key: [name: :id, type: :binary_id],
```
simonprev, line 16: I see that in other (private) repo of our organization we have:

```elixir
migration_primary_key: [type: :binary_id],
migration_primary_key: [name: :id, type: :uuid, default: {:fragment, "gen_random_uuid()"}]
migration_primary_key: [name: :id, type: :binary_id, default: {:fragment, "gen_random_uuid()"}]
```

and you are proposing `migration_primary_key: [name: :id, type: :binary_id],`

I think that `binary_id` is the documented one but I think we should add the default. It makes it easier to create row in a GUI like TablePlus 😉

lib/elixir_boilerplate/schema.ex
```diff
@@ -0,0 +1,15 @@
...
+    quote do
+      use Ecto.Schema
+
+      import Ecto.Changeset
+
+      alias Ecto.{Schema, UUID}
+
+      @primary_key {:id, :binary_id, autogenerate: true}
+      @foreign_key_type :binary_id
+      @timestamps_opts [type: :utc_datetime_usec]
```
simonprev, line 12: We also have `migration_timestamps: [type: :utc_datetime_usec]` in the config for a lot of our app.

simonprev (approved):

PR #396, opened 2023-09-18
Remove DEBUG_ERRORS env var and replace it to true in dev, false in prod

simonprev:
## 📖 Description

Two reasons:

- It’s weird to expose a way in production to see debug errors (potentially leaking sensitive informations)
- It seems that the value is somehow cached in the compile process since changing the value in the env do not change the value when we start the app. This result in a poor experience when all we want to to is set `debug_errors:` to false in dev.

BONUS: While debugging this issue I copied a line we have on another project: https://hexdocs.pm/phoenix/Phoenix.html#plug_init_mode/0

> Returns the :plug_init_mode that controls when plugs are initialized.
> We recommend to set it to :runtime in development for compilation time improvements. It must be :compile in production (the default).

## 🦀 Dispatch

- `#dispatch/elixir`

### DIFF
--- a/.env.dev
+++ b/.env.dev
@@ -12,7 +12,6 @@
 
 # Server configuration
 CORS_ALLOWED_ORIGINS=*
-DEBUG_ERRORS=true
 PORT=4000
 SECRET_KEY_BASE= # Generate secret with `mix phx.gen.secret`
 SESSION_KEY=elixir_boilerplate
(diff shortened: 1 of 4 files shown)


charlesdemers (approved):

remi:
So if we want to test our custom error pages in development, we just need to temporarily set `debug_errors: false` in `config/dev.exs`, right?

gcauchon (commented):
👌🏻

garno (commented):
Late 👍

PR #457, opened 2023-12-06
Halt server if migrations fail on boot

remi:
## 📖 Description

It doesn’t make much sense to start the server if the migrations (that are run on container startup) fail.

## 📝 Notes

I also updated a few lines based on the [`mix phx.gen.release --docker`](https://github.com/phoenixframework/phoenix/tree/main/priv/templates/phx.gen.release) template.

## 🦀 Dispatch

- `#dispatch/elixir`

### DIFF
--- a/rel/overlays/bin/server
+++ b/rel/overlays/bin/server
@@ -1,7 +1,3 @@
 #!/bin/sh
 cd -P -- "$(dirname -- "$0")"
-
-# Always run migrations
-./elixir_boilerplate eval ElixirBoilerplate.Release.migrate
-
-exec ./elixir_boilerplate start
+PHX_SERVER=true exec ./elixir_boilerplate start
--- a/Dockerfile
+++ b/Dockerfile
@@ -79,12 +79,12 @@ ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US:en
 ENV LC_ALL en_US.UTF-8
 
-WORKDIR "/opt/elixir_boilerplate"
-RUN chown nobody /opt/elixir_boilerplate
+WORKDIR "/app"
+RUN chown nobody /app
 
 # Only copy the final release from the build stage
 COPY --from=otp-builder --chown=nobody:root /app/_build/prod/rel/elixir_boilerplate ./
 
 USER nobody
 
-CMD ["/opt/elixir_boilerplate/bin/server"]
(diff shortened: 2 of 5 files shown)


lewazo (approved):

rel/overlays/bin/server
```diff
@@ -1,7 +1,3 @@
 #!/bin/sh
 cd -P -- "$(dirname -- "$0")"
-
-# Always run migrations
-./elixir_boilerplate eval ElixirBoilerplate.Release.migrate
-
-exec ./elixir_boilerplate start
+PHX_SERVER=true exec ./elixir_boilerplate start
```
JoeyBG, line 3: Si tu mets ça ici je pense qu'on peut enlever le `server: true` dans `prod.exs`

Boubalou (approved):

rel/overlays/bin/server
remi: Yup, I moved it from `config/prod.exs` to `config/runtime.exs`.

gcauchon: Do we need to add à default value to `PHX_SERVER` in `.env.dev` or `get_env/2` returning `nil` works?

remi: Oh you’re right 🤔 I looked at what Phoenix does and it looks like there’s no server running in development 😳

remi: Yes I know (I added a similar line as well), but `PHX_SERVER` is still not set when you run `mix phx.server` in development, isn’t it? 🤔

The only way to set `server: true` is to explicitely use the `PHX_SERVER` environment variable: https://github.com/search?q=repo%3Aphoenixframework%2Fphoenix+%22server%3A+true%22&type=code

remi: AH!

> `:server` - when true, starts the web server when the endpoint supervision tree starts. Defaults to false. The [mix phx.server](https://hexdocs.pm/phoenix/1.7.10/Mix.Tasks.Phx.Server.html) task automatically sets this to true
>
>  — [_reference_](https://hexdocs.pm/phoenix/1.7.10/Phoenix.Endpoint.html#module-runtime-configuration)

So setting `server: false` overrides `mix phx.server` behavior.

IMMA FIX IT!

PR #458, opened 2023-12-07
Restrict graphql content types

matkev:
## 📖 Description

Restrict the allowed graphql content-types to only `application/json`

Also, include subdomains in the HSTS header as recommended by Escape, to apply the Strict-Transport-Security rule to the site's subdomains as well.

## 🦀 Dispatch

- `#dispatch/elixir`

### DIFF
--- a/lib/elixir_boilerplate_web/endpoint.ex
+++ b/lib/elixir_boilerplate_web/endpoint.ex
@@ -4,7 +4,7 @@ defmodule ElixirBoilerplateWeb.Endpoint do
 
   alias Plug.Conn
 
-  @plug_ssl Plug.SSL.init(rewrite_on: [:x_forwarded_proto])
+  @plug_ssl Plug.SSL.init(rewrite_on: [:x_forwarded_proto], subdomains: true)
 
   socket("/socket", ElixirBoilerplateWeb.Socket)
   socket("/live", Phoenix.LiveView.Socket, websocket: [connect_info: [session: {ElixirBoilerplateWeb.Session, :config, []}]])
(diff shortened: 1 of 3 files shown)


lucbernard (approved):

lib/elixir_boilerplate_web/endpoint.ex
```diff
@@ -4,7 +4,7 @@ defmodule ElixirBoilerplateWeb.Endpoint do
 
   alias Plug.Conn
 
-  @plug_ssl Plug.SSL.init(rewrite_on: [:x_forwarded_proto])
+  @plug_ssl Plug.SSL.init(rewrite_on: [:x_forwarded_proto], subdomains: true)
```
gcauchon, line 7: Why are we using a module attribute to initialize the SSL plug? We should do it the same way we are initializing other plugs in the endpoint?

https://github.com/mirego/elixir-boilerplate/blob/a4fe43ce3ea224b5c7b03488bfb9e56d8dc52b09/lib/elixir_boilerplate_web/endpoint.ex#L93-L123

```elixir
defp force_ssl(conn, _opts) do
  if Application.get_env(:elixir_boilerplate, :force_ssl) do
    opts = Plug.SSL.init(rewrite_on: [:x_forwarded_proto], subdomains: true)

    Plug.SSL.call(conn, opts)
  else
    conn
  end
end
```

matkev: It looks like it used to be initialized that way, but was changed in #130 to use a module attribute since it does not depend on any runtime environment variables.

remi: Yep, nothing needs to change here. 

> Also, include subdomains in the HSTS header as recommended by Escape, to apply the Strict-Transport-Security rule to the site's subdomains as well.

This should be done in a separate pull request.

PR #499, opened 2024-02-15
Sort keys in format_bindings/1

voughtdq:
## 📖 Description

Tests were failing when the left side's key order differed from the right side's. Explicitly ordering the keys makes the test less flaky. 

## 📝 Notes

Added a doctest to test the ordering.

## 📓 References

N/A

## 🦀 Dispatch

- `#dispatch/elixir`


remi (approved):
Good catch! 👌

PR #504, opened 2024-02-21
Revert "Restrict graphql content types (#458)"

### LINKED ISSUE #34: Add Absinthe GraphQL tooling
We have been using [GraphQL](https://graphql.org/) on every web projects for months, I think we should include the [Absinthe](https://absinthe-graphql.org/) setup in the _boilerplate_ to simplify scaffolding of upcoming Elixir projects.

MathieuLegault1:
This reverts commit 6a4173ac3f106fc68a069905b0fb84191020e9ff.

## 📖 Description

We unfortunately cannot remove the `Plug.Parser` from endpoint since the plug `Plug.MethodOverride` requires the body to be parsed before calling it. I've tried to move the plug `MethodOverride` to `router.ex` but unfortunately it doesn't work since a pipeline is only executed if a route matches. 

In the documentation:
```
Note that router pipelines are only invoked after a route is found. No plug is invoked in case no matches were found.
```

For now we revert this commit and we can check after what is the best way to restrict the GraphQL endpoint to only json.

## 🦀 Dispatch

- `#dispatch/elixir`

### DIFF
--- a/lib/elixir_boilerplate_graphql/router.ex
+++ b/lib/elixir_boilerplate_graphql/router.ex
@@ -5,13 +5,6 @@ defmodule ElixirBoilerplateGraphQL.Router do
     @moduledoc false
     use Plug.Router
 
-    plug(
-      Plug.Parsers,
-      parsers: [:json],
-      pass: [],
-      json_decoder: Phoenix.json_library()
-    )
-
     plug(:match)
     plug(:dispatch)
 
--- a/lib/elixir_boilerplate_web/endpoint.ex
+++ b/lib/elixir_boilerplate_web/endpoint.ex
@@ -40,6 +40,13 @@ defmodule ElixirBoilerplateWeb.Endpoint do
   plug(Plug.RequestId)
   plug(Plug.Telemetry, event_prefix: [:phoenix, :endpoint])
 
+  plug(
+    Plug.Parsers,
+    parsers: [:urlencoded, :multipart, :json],
+    pass: ["*/*"],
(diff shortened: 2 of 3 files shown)


gcauchon:
I personnaly challenge the approach to break routing into multiple modules… It does not bring any value, and we even have to add a plug to overcome undesired side effects!

https://github.com/mirego/elixir-boilerplate/blob/main/lib/elixir_boilerplate_web/endpoint.ex#L53

In the large PR a did a few months ago to _re-sync_ our boilerplate with the community approach, it was actually one of the first clean-up I did… Too bad it was never merged 🤷🏻‍♂️ 

https://github.com/mirego/elixir-boilerplate/pull/249

remi (approved):

lib/elixir_boilerplate_graphql/router.ex
```diff
@@ -5,13 +5,6 @@ defmodule ElixirBoilerplateGraphQL.Router do
     @moduledoc false
     use Plug.Router
 
-    plug(
-      Plug.Parsers,
-      parsers: [:json],
-      pass: [],
-      json_decoder: Phoenix.json_library()
-    )
-
```
remi, line 14: @MathieuLegault1 But we could add a custom plug here to _make sure_ the request content-type is `application/json` right?

MathieuLegault1: Yes we could. I'll merge this PR so nobody has this problem if someone forks it and I'll look to do it this week.

PR #564, opened 2025-11-18
Enable observer in dev

Madumo:
I used [the same pattern phoenix does](https://github.com/phoenixframework/phoenix/blob/v1.8.1/mix.exs#L62-L68)


simonprev (approved):
I like it!

